PR #1439, opened 2015-01-04
1405 output proof details

eugenk:
Shall fix #1405.
## Example

Running
```
curl http://localhost:8000/prove/file%3A%2F%2F%2Ftmp%2FGroups.casl
```
with this `/tmp/Groups.casl`
```
spec Group =
  sort s
  ops 0:s;
      -__ :s->s;
      __+__ :s*s->s, assoc
  forall x,y:s
  . x+(-x) = 0
  . x+0=x %(leftunit)%
  . 0+x=x %(rightunit)% %implied
  . 0+0=0 %(zero_plus)% %implied
end

spec Empty =
  sort t
end
```
results in the following xml response
```xml
<?xml version='1.0' ?>
<Results>
  <Group>
    <results>
      <goal>
        <name>rightunit</name>
        <result>Proved</result>
        <details></details>
        <usedProver>SPASS</usedProver>
        <tacticScript>
          <timeLimit>1</timeLimit>
          <extraOpts>
            <option>-DocProof</option>
          </extraOpts>
        </tacticScript>
        <proofTree></proofTree>
        <usedTime>
          <seconds>0</seconds>
          <components>
            <hours>0</hours>
            <minutes>0</minutes>
            <seconds>0.000000000000</seconds>
          </components>
        </usedTime>
        <usedAxioms>
          <axiom>leftunit</axiom>
          <axiom>Ax2</axiom>
          <axiom>ga_assoc___+__</axiom>
        </usedAxioms>
        <proverOutput><![CDATA[
--------------------------SPASS-START-----------------------------
Input Problem:
1[0:Inp] ||  -> equal(o__Plus__(U,zero),U)**.
2[0:Inp] || equal(o__Plus__(zero,skc1),skc1)** -> .
3[0:Inp] ||  -> equal(o__Plus__(U,minus__(U)),zero)**.
4[0:Inp] ||  -> equal(o__Plus__(o__Plus__(U,V),W),o__Plus__(U,o__Plus__(V,W)))**.
 This is a unit equality problem.
 This is a problem that has, if any, a non-trivial domain model.
 The conjecture is ground.
 Axiom clauses: 3 Conjecture clauses: 1
 Inferences: IEqR=1 ISpR=1 ISpL=1
 Reductions: RFRew=1 RBRew=1 RFMRR=1 RBMRR=1 RObv=1 RUnC=1 RTaut=1 RFSub=1 RBSub=1
[...]

PR #1442, opened 2015-01-09
Add missing proofLines definitions.

eugenk:
In #1439 I forgot to set the `proofLines` in two files because the compiler didn't complain (I already had up to date object files for them). Now the compiler shouldn't complain about any missing `proofLines`.

### DIFF
--- a/OWL2/ProveFact.hs
+++ b/OWL2/ProveFact.hs
@@ -240,11 +240,6 @@ runFact sps cfg saveFact thName nGoal = do
         { tacticScript = tScript }
     disProvedStatus = defaultProofStatus {goalStatus = Disproved}
     provedStatus ut =
-      ProofStatus { goalName = senAttr nGoal
-                  , goalStatus = Proved True
-                  , usedAxioms = []
-                  , usedProver = proverName factProver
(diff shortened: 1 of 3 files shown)


cmaeder:
instead of using the bare constructor `ProofStatus` to construct a new record it is usually better to (re-) use a function like openProofStatus that sets appropriate default values for fields that are newly added or only used for special cases.

OWL2/ProveFact.hs
```diff
@@ -246,4 +246,5 @@ runFact sps cfg saveFact thName nGoal = do
                   , usedProver = proverName factProver
                   , proofTree = emptyProofTree
                   , usedTime = ut
-                  , tacticScript = tScript }
+                  , tacticScript = tScript
```
cmaeder: here defaultProofStatus should be reused for provedStatus as is done for disProvedStatus

cmaeder:
Travis still contains a warning for `SoftFOL/ProveHyperHyper.hs:178:24`

eugenk:
Done with 070dd55

cmaeder:
:+1:

PR #1443, opened 2015-01-10
1441 parse json data from request body

### LINKED ISSUE #1441: parse JSON data from request body
Currently, the RESTful API only uses form data. As pointed out in an ontohub developer meeting, the API should parse JSON data in a non-GET request.

eugenk:
This shall fix #1441.

Form data is still supported, but not encouraged.
JSON request bodies are parsed with [aeson](http://hackage.haskell.org/package/aeson-0.8.0.2/docs/Data-Aeson.html), if the content type is set to `application/json`. Otherwise form data is assumed.

Now, aeson must be installed with cabal prior to compiling hets.

You can try the current behaviour with these calls:

* POST with Form data
```
curl -F theorems='["rightunit"]' -F axioms='["Ax2","ga_assoc___+__","leftunit"]' -F node=Group http://localhost:8000/prove/file%3A%2F%2F%2Fprivate%2Ftmp%2FGroups.casl
```
* POST with JSON
```
curl -X POST -H "Content-Type: application/json" -d '{"theorems": ["rightunit"], "axioms": ["Ax2","ga_assoc___+__","leftunit"], "node": "Group"}' http://localhost:8000/prove/file%3A%2F%2F%2Ftmp%2FGroups.casl
```
* GET with query string
```
curl http://localhost:8000/prove/file%3A%2F%2F%2Ftmp%2FGroups.casl\?theorems\=rightunit\;axioms\=Ax2,ga_assoc___%2B__,leftunit\;node\=Group
```
while having the `/tmp/Groups.casl`:
```casl
spec Group =
  sort s
  ops 0:s;
      -__ :s->s;
      __+__ :s*s->s, assoc
  forall x,y:s
  . x+(-x) = 0
  . x+0=x %(leftunit)%
  . 0+x=x %(rightunit)% %implied
  . 0+0=0 %(zero_plus)% %implied
end

spec Empty =
  sort t
end
```

### DIFF
--- a/Common/Json.hs
+++ b/Common/Json.hs
@@ -13,7 +13,7 @@ inspired by Yuriy Iskra's json2-types hackage package
 -}
 
 module Common.Json
-  ( Json
+  ( Json (..)
   , ppJson
   , mkJStr
   , mkJBool
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -182,7 +182,7 @@ hetsServer opts1 = do
   run port $ \ re -> do
    let respond = liftIO . return
 #endif
-   requestParams <- parseRequestParams re
+   requestBodyParams <- parseRequestParams re
    let rhost = shows (remoteHost re) "\n"
        ip = getIP4 rhost
        white = matchWhite ip wl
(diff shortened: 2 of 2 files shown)


cmaeder:
I've tried to avoid new dependencies like aeson (that pull in even more deps) as much as possible. There also is a JSON parser pJson in Common.Json. (It took me ages to get rid of HXT, a fourth XML library beside HaXml, hexpat and xml-light.) 

But I've no time to contrib much to Hets currently.

cmaeder:
The deps must be also added when builded ubuntu packages. 
https://github.com/spechub/Hets/blob/master/utils/debian/auto-package/debian-common/control
https://github.com/spechub/Hets/blob/master/utils/debian/auto-package/debian-server/control

cmaeder:
I usually installed the packages using https://github.com/spechub/Hets/blob/master/utils/installHetsPkgs.sh.

eugenk:
Now there is no more `Aeson` dependency. I removed it in favor of `Common.Json`. The commits mentioning `Aeson` are removed by rebase. I think this branch is done.

PR #1444, opened 2015-01-14
1384 output prove result as json

eugenk:
This branch introduces a bit of output handling for the RESTful server. It is built on top of the branch `1441-parse_json_data_from_request_body` because I did'n want to get conflicts once 1441 is merged.

### Feature list
* **new**: JSON output for proofs (main feature)
* **new** Parameters to hide some output of the `prove` command (details and the proof itself)
* **enhanced**: XML output for proofs uses the same structure as the JSON output.
* **enhanced**: Generic conversion of ordinary Haskell data types to JSON and XML: Now the serialization doesn't leak much of the Haskell specific part of the internal data structures, i.e. type constructors are not shown in every single element, `Nothing`s are left out. `Maybe a` and `Either a b` are simply considered as their values without the constructors.
* **enhanced**: A little bit of modularization of the server: I think we will add more structured JSON and XML (and maybe other formats as well) output to the server in the future, i.e. to more commands, and the new `PGIP.Output.*` module structure is supposed help the developers to oversee the project.

### Calls and Output
Hets yields [this example output](https://gist.github.com/eugenk/452bbcd7c9b928e221bb) for the calls:
```
curl -X POST -H "Content-Type: application/json" -d '{"format": "xml"}' http://localhost:8000/prove/file%3A%2F%2F%2Fprivate%2Ftmp%2FGroups.casl
```
```
curl -X POST -H "Content-Type: application/json" -d '{"format": "json"}' http://localhost:8000/prove/file%3A%2F%2F%2Fprivate%2Ftmp%2FGroups.casl
```

### Comments
The new output is *only* used by the RESTful API. The old webAPI still has its old output. I didn't want to break compatibility.

I recommend to review this pull request commit-by-commit because it's quite a big one and the commits don't revert much of the previous ones.


Common/Utils.hs
```diff
@@ -26,6 +26,7 @@ module Common.Utils
   , nubOrdOn
   , atMaybe
   , readMaybe
+  , orElse
```
cmaeder: only used once, so not really a utility (yet).

Common/Utils.hs
```diff
@@ -184,6 +185,13 @@ readMaybe s = case filter (all isSpace . snd) $ reads s of
   [(a, _)] -> Just a
   _ -> Nothing
 
+-- | Left-biased choice on maybes
+-- Extracted from Data.Generics.Aliases (syb-0.4.4)
+orElse :: Maybe a -> Maybe a -> Maybe a
+x `orElse` y = case x of
```
cmaeder: I'm no friend of using alphanumeric identifiers as infix operators. 
A shorter implementation would be:
`orElse x y = maybe y Just x`

PGIP/Output.hs
```diff
@@ -0,0 +1,6 @@
+module PGIP.Output
```
cmaeder: I'm no friend of re-exports. In a long list of imports two direct imports will not harm.

Common/Utils.hs
eugenk: Correct, but it Utils is the only place it really fits in.

cmaeder:
You surely do not need to follow my taste, but I used to run  https://github.com/spechub/Hets/blob/master/utils/checkSource.sh over my files. For Common.Utils I get:

```
maeder@Opti:~/Hets$ checkSource.sh Common/Utils.hs
Common/Utils.hs:189:1: join (or separate) consecutive comments
Common/Utils.hs:191:14: Error: Use mplus
Found:
  maybe y Just x
Why not:
  mplus x y
```

So maybe orElse is not needed. (Try checkSource.sh for your other files!)

eugenk:
You're right. `mplus` does exactly what I need.

cmaeder:
I guess this could be merged. :+1:

cmaeder:
Hang on, why do you rely on Parsec.Prim? Hets should compile with parsec1 only.

```
PGIP/Server.hs:70:8:
    Could not find module ‘Text.Parsec.Prim’
    It is a member of the hidden package ‘parsec-3.1.7’.
```

cmaeder:
Hi, I've changed the import (1 line), sorry. Does it still work for you?

eugenk:
Yes, it still works.

eugenk:
Can this be merged now?

cmaeder:
yes :+1:

cmaeder:
@eugenk I've noticed several warnings (on pollux and travis) in `OWL2/ProveFact.hs:243:7`, ` SoftFOL/ProveHyperHyper.hs:178:24`, `QBF/ProveDepQBF.hs:156:13`

```
Fields of ‘ProofStatus’ not initialised: proofLines
```

These turn up after full recompilation and should be removed.

eugenk:
See #1442 for this

PR #1446, opened 2015-01-29
Fix file upload

eugenk:
The call of `parseRequestParams` consumed the request body, resulting
in the old webAPI not having anything from body to parse.

The call is simply moved to a place where only the RESTful API is used.


cmaeder:
:+1: merge if travis does not complain

PR #1447, opened 2015-01-30
Support Warp 1 and 2.

eugenk:
This allows to compile Hets with older GHC/warp versions.

### DIFF
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -265,11 +265,7 @@ parseRequestParams request =
           jsonStringOrArray str =
             if B8.head str == '[' then B8.unpack str else show str
       in do
-#ifdef WARP3
         (formDataB8, _) <- parseRequestBody lbsBackEnd request
-#else
-        (formDataB8, _) <- parseRequestBody lbsBackEnd request
-#endif
         return $ parseJson $ toJsonObject formDataB8
 
 #ifdef WARP1
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -161,11 +161,7 @@ type ResIO a = IO a
 #ifdef WARP3
 type WebResponse = (Response -> IO ResponseReceived) -> IO ResponseReceived
 #else
-#ifdef WARP1
-type WebResponse = (Response -> ResourceT IO Response) -> ResourceT IO Response
-#else
-type WebResponse = (Response -> IO Response) -> IO Response
-#endif
+type WebResponse = (Response -> ResIO Response) -> ResIO Response
 #endif
 
 hetsServer :: HetcatsOpts -> IO ()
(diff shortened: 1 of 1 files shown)


PGIP/Server.hs
```diff
@@ -230,8 +232,11 @@ hetsServer opts1 = do
            -- only otherwise stick to the old response methods
            else oldWebApi newOpts tempLib permFile sessRef re pathBits qr2
              meth respond
-
+#ifdef WARP1
```
cmaeder: you should make a conditional "type ResIO a = IO a" (or "ResourceT IO a") and also use it within "type  WebResponse" (about line 158). This would avoid the #ifdefs for the type signatures.

PGIP/Server.hs
```diff
@@ -245,19 +250,51 @@ parseRequestParams request =
...
             ++ intercalate ", " (map (\ (k, v) ->
                   show k ++ ": " ++ jsonStringOrArray v) assocList)
             ++ "}"
           jsonStringOrArray str =
             if B8.head str == '[' then B8.unpack str else show str
       in do
+#ifdef WARP3
+        (formDataB8, _) <- parseRequestBody lbsBackEnd request
+#else
         (formDataB8, _) <- parseRequestBody lbsBackEnd request
```
cmaeder: here I see not difference in the cases.

PGIP/Server.hs
```diff
@@ -153,11 +155,14 @@ matchWhite ip l = null l || any (matchIP4 ip) l
...
+type ResIO a = IO a
 #else
 #ifdef WARP1
 type WebResponse = (Response -> ResourceT IO Response) -> ResourceT IO Response
+type ResIO a = ResourceT IO a
 #else
 type WebResponse = (Response -> IO Response) -> IO Response
+type ResIO a = IO a
 #endif
 #endif
```
cmaeder: this should be:

```
#ifdef WARP1
type ResIO a = ResourceT IO a
#else 
type ResIO a = IO a
#endif
#ifdef WARP3
type WebResponse = (Response -> IO ResponseReceived) -> IO ResponseReceived
#else
type WebResponse = (Response -> ResIO Response) -> ResIO Response
#endif
```

PGIP/Server.hs
```diff
@@ -256,8 +262,22 @@ parseRequestParams request =
...
+    receivedRequestBody :: ResIO B8.ByteString
+#ifdef WARP3
+    receivedRequestBody = requestBody request
+#else
+#ifdef WARP1
+    receivedRequestBody = liftM (B8.pack . BS.unpack) $ lazyRequestBody request
+    lazyRequestBody :: Request -> ResourceT IO BS.ByteString
+    lazyRequestBody = fmap BS.fromChunks . lazyConsume . requestBody
+#else
+    receivedRequestBody = liftM (B8.pack . BS.unpack) $ lazyRequestBody request
```
cmaeder: here is also a duplicate receivedRequestBody definition.

cmaeder:
:+1:

PR #1449, opened 2015-02-13
check only if unexpanded IRIs are predefined

### LINKED ISSUE #1431: OWL fails with undeclared `ObjectProperty: rdfs:Literal`
```
hets -v2 https://ontohub.org/test-ifc/IFCfragment.owl
```

cmaeder:
This is supposed to fix #1431. Some tests should be made if other OWL examples are uneffected by this change. Could someone review this? 

XML creates expanded IRIs without a prefix.
"function Expand" checked for a predefined local part
(and a missing or matching prefix) and inserted a missing
prefix (which caused the problem of #1431).

This patch now only considers unexpanded IRIs as predefined.

Another minor change ensures that a prefix is always only set,
when it was empty before.
This was not case for the rdfs annotation properties:
label, comment, seeAlso, isDefinedBy

### DIFF
--- a/OWL2/AS.hs
+++ b/OWL2/AS.hs
@@ -294,9 +294,8 @@ preDefMaps sl pref = let
  in (sl, pref, sp)
 
 checkPredefAux :: PreDefMaps -> IRI -> Maybe (String, String)
-checkPredefAux (sl, pref, exPref) u = case expandedIRI u of
-  "" -> Nothing
-  _ -> let lp = localPart u in case namePrefix u of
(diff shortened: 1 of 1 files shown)


mcodescu:
looks fine to me :+1:

mcodescu:
I will try some OWL ontologies now.

cmaeder:
I had to change more. Basically .omn files as input or output might be affected.

cmaeder:
I've tested some bioportal stuff and it looks fine to me. I'll merge this, if someone agrees or nobody complains

PR #1453, opened 2015-02-18
Ghc 7.10

cmaeder:
The following changes should be added to ease transition to ghc-7.10. They should not harm for older ghc versions. I've run a test using ghc-7.10RC2, but hexpat, wai-extra, warp und gtk stuff were still missing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,20 @@
-language: haskell
+env:
+ - GHCVER=7.8.4 CABALVER=1.20
 before_install:
+ - travis_retry sudo add-apt-repository -y ppa:hvr/ghc
+ - travis_retry sudo apt-get update
+ - travis_retry sudo apt-get install cabal-install-$CABALVER ghc-$GHCVER happy-1.19.4 alex-3.1.3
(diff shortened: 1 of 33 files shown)


eugenk:
You seem to have missed one instance in https://github.com/spechub/Hets/blob/ghc-7.10/Common/Lib/Maybe.hs#L23
```
[ 10 of 753] Compiling Common.Lib.Maybe ( Common/Lib/Maybe.hs, Common/Lib/Maybe.o )

Common/Lib/Maybe.hs:23:10: Warning:
    ‘MaybeT’ is an instance of Monad but not Applicative - this will become an error in GHC 7.10, under the Applicative-Monad Proposal.
```

cmaeder:
Right, the files  HolLight/HolLight2DG.hs and Common/SAX.hs that use Common/Lib/Maybe.hs have not yet been compiled with ghc-7.10-RC2 because hexpat was not available (and is "ifdef"-ed in hets.hs)

eugenk:
Alright, so this shall be done in another branch? In that case: :+1:

cmaeder:
no, I'll check hexpat this week.

eugenk:
:+1:

PR #1456, opened 2015-02-20
1455 owl predefs

### LINKED ISSUE #1455: unknown Datatype: 'integer' - Hets fails analyzing DOL ontology
```bash
hets https://github.com/ontohub/ontohub/raw/staging/spec/fixtures/ontologies/dol/double_mapped_logic_translated_blendoid.dol
```
fails with
```
*** Error:
unknown Datatype: 'integer'
```
and tests on ontohub fail because of this issue.

cmaeder:
fix #1455 (that was caused by fixing #1431)

### DIFF
--- a/OWL2/AS.hs
+++ b/OWL2/AS.hs
@@ -91,7 +91,7 @@ unamedS :: String
 unamedS = "//www." ++ dnamedS
 
 dnamedS :: String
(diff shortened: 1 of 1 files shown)


eugenk:
:+1:

PR #1460, opened 2015-03-05
avoid rebuilding Isabelle/IsaExport.hs

cmaeder:
the binary DtdToHaskell comes with HaXml from hackage, but we compile modified sources (with many warning), because we only want to create a qualified module name inside the created haskell module. (An early versions of HaXml was also buggy.)

If the dtd (Isabelle/IsaExport.dtd) actually changes the file Isabelle/IsaExport.hs needs to be regenerated manually now.

This change just saves about 3 seconds re-build time (and a couple of ugly output lines)


cmaeder:
I'm not sure if this is really worth merging. It could be closed (if it annoys as open pull request)

PR #1461, opened 2015-03-05
try cabal update

cmaeder:
I've only created this pull request to see the travis builds. Is there a different way?

Also blindly comitting the .travis.yml file is also no good way to debug travis problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_install:
  - sudo apt-get install -qq zlib1g-dev libncurses-dev libcairo2-dev libglib2.0-dev libpango1.0-dev libgtk2.0-dev libglade2-dev
 install:
  - travis_retry cabal update
+ - cabal install alex happy
  - cabal install gtk2hs-buildtools
  - cabal install glade parsec1
  - cabal install --only-dependencies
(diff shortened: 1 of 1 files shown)


cmaeder:
"git rebase master; git pull; git push" caused the above history, which looks confusing to me. Do I something wrong (for this trivial change)? Should I've simply started a new branch?

eugenk:
After rebase, the local branch and the remote (github) branch usually diverge. If you pull then, the remote one gets merged into the local one. And the push (I assume you force-pushed) overwrites the remote branch.

Usually you do
```bash
git checkout master
git pull
git checkout feature-branch
git rebase master
git push -f
```
to push the clean rebased branch to the remote.

Note that this overwrites the history of the branch and everyone who already has pulled it needs to `git reset --hard origin/feature-branch` to get the changes. This also causes them to lose their own changes.

In such feature branches, this should not be a problem. Rebase inside of master will make the other developers angry though ;)

eugenk:
To fix the history here, you can do:
``` bash
git reset --hard master
git cherry-pick 12db4b8
git cherry-pick e654b3c
git cherry-pick 071687e
git push -f
```

eugenk:
I just made the suggested steps to fix the history.

eugenk:
:+1:

cmaeder:
Only one PR of this one and #1462 make sense. Which one? This one takes longer but could test more configurations.

eugenk:
I think time should not be an issue, so I prefer this one because testing more configurations ist usually good.

PR #1466, opened 2015-03-10
1203 owl2casl translation

mcodescu:
This makes Fabian's file go through, as the prefixes are now expanded before translation of symbols to CASL. Also all the errors reported in 1203 are fixed. I will test if sentence translation contains other bugs, but I can create a new issue if that's going to be the case.


tillmo:
after discussing in person with Christian: :+1:

PR #1474, opened 2015-03-24
Remove ExtModal2OWL from the list of comorphisms.

### LINKED ISSUE #1471: Too many provers are shown
When analysing the spec
```
spec strict_partial_order =
  sort elem
  pred __<__ : elem*elem
  forall x,y,z:elem
  . not x < x %(irreflexive)%
  . x < y /\ y < z => x <z %(transitive)%
  . x < y => not y < x  %(asymmetric)%  %implied
end
```
and then trying to prove, I can also select the truth tables prover, but then (no wonder) get an error
```
*** Error:
not a propositional formula: forall x : elem
. not (pred __<__ : elem * elem)((var x : elem), (var x : elem))
```
I also got Fact and Pellet as provers (with the same error message), but cannot reproduce that now. But I can reproduce it using the prover selection of ontohub.

eugenk:
This shall fix #1471. I can't test the result because I don't have the needed provers. Please test if this really fixes #1471.

### DIFF
--- a/Comorphisms/CASL2Prop.hs
+++ b/Comorphisms/CASL2Prop.hs
@@ -84,7 +84,7 @@ instance Comorphism CASL2Prop
       map_symbol CASL2Prop _ = mapSym
       map_sentence CASL2Prop _ = trForm
       map_morphism CASL2Prop = mapMor
-      has_model_expansion CASL2Prop = True
+      has_model_expansion CASL2Prop = False
       is_weakly_amalgamable CASL2Prop = True
       isInclusionComorphism CASL2Prop = False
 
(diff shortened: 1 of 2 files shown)


cmaeder:
The proper fix would be to specify the sublogic for which this comorphism works. The code was used in the SHIP project. Could someone try to set "has_model_expansion" in Comorphisms.ExtModal2CASL to false and see if the comorphism is still considered?
(Otherwise, unused/unmainted code should be deleted from the repository.)

tillmo:
Acutally, removing ExtModal2CASL was the wrong idea. CASL2Prop was the evil guy. See 5b65f67

tillmo:
:+1: (if you give me :+1: for my commit)

eugenk:
:+1:, however, I can't test this, because I don't get `truth tables` as a prover in the provers list (as stated in the issue)

PR #1476, opened 2015-03-26
Use consistent naming of provers

### LINKED ISSUE #1475: Use consistent naming of provers (RESTful API)
The commands `provers` and `prove` use different naming for the same provers. They need to have the same names for ontohub's proving feature to work properly.

eugenk:
This shall fix #1475.

Instead of the prover set in the ProofStatus object, we use the Prover
set in the proving/consistency-checking functions. This allows us to
be sure to have the prover object and use the same name-retrieval
functions as in the provers/consistency-checkers command.

The `usedProver` is like an id for ontohub and we need it to be the same all over the responses.

Running
```
curl -X POST -H 'Content-Type: application/json' http://localhost:8000/prove/http%3A%2F%2Fdevelop.ontohub.org%2Fdefault%2Fpartial_order%2F%2Fstrict_partial_order/auto -d '{"format":"json", "node":"strict_partial_order", "theorems": ["asymmetric"], "prover":"MathServeBroker"}'
```

previously we got:
```json
[{
"node": "strict_partial_order",
"goals": [{
 ... "usedProver": "MathServe Broker [via MathServe]", ...
 }]
}]
```

now we get:
```json
[{
"node": "strict_partial_order",
"goals": [{
 ... "usedProver": "MathServeBroker", ...
 }]
}]
```

while the list of provers remains:
```json
{
"provers": [
 "QuickCheck",
 "Pellet",
 "eprover",
 "darwin",
 "darwin-non-fd",
 "Vampire",
 "MathServeBroker",
 "SPASS"]
}
```

### DIFF
--- a/PGIP/Output/Proof.hs
+++ b/PGIP/Output/Proof.hs
@@ -10,6 +10,8 @@ module PGIP.Output.Proof
   , formatProofs
   ) where
 
+import PGIP.Common
+
 import PGIP.Output.Formatting
 import PGIP.Output.Mime
 
@@ -30,8 +32,8 @@ import Numeric
 
 import Text.XML.Light (ppTopElement)
 
-type ProofResult = (String, String, String,
-                -- (goalName, goalResult, goalDetails
+type ProofResult = (String, String, String, ProverOrConsChecker,
+                -- ^(goalName, goalResult, goalDetails, prover
                     AnyComorphism, Maybe (LP.ProofStatus G_proof_tree))
                 -- translation, proofStatusM)
 type ProofFormatter =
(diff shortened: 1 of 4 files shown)


eugenk:
I just rebased and force-pushed because I modularized this a little more for future features.

PGIP/Output/Proof.hs
```diff
@@ -30,8 +32,8 @@ import Numeric
 
 import Text.XML.Light (ppTopElement)
 
-type ProofResult = (String, String, String,
-                -- ^(goalName, goalResult, goalDetails
+type ProofResult = (String, String, String, ProverOrConsChecker,
+                -- ^(goalName, goalResult, goalDetails, prover
```
cmaeder: there are wrong haddock markers (check using "make doc"). I had to fix them in https://github.com/spechub/Hets/tree/ignazio1977-master

eugenk: Thanks. Fixed in cf47588031e531a4808ec3bf221a02c62ab51b44

tillmo:
:+1:

PR #1486, opened 2015-04-08
Always update version.

### LINKED ISSUE #1485: Update version on each make-run
Running `make` only updates the hets version if there was no `rev.txt`. Delete the `rev.txt` each time when executing the default target.

eugenk:
Shall fix #1485. This makes `Driver.Version` compile always and thus set the current version.


cmaeder:
If it works this way then it is fine. I used to copy the rev.txt dependency to the all target to force its regeneration. :+1:

eugenk:
For me it worked perfectly fine.

PR #1489, opened 2015-04-20
1488 put json keys in snake case

### LINKED ISSUE #1488: Put JSON keys in snake_case
The convention in JSON is to have snake_case keys in objects. Convert camelCase to snake_case in the automatic JSON generator.

eugenk:
This shall fix #1488.

Truncated example output:

```json
❯ http --json http://localhost:8000/prove/http%3A%2F%2Fstaging.ontohub.org%2Fsandbox%2FSimple_Implications/auto axioms:='["leftunit"]' theorems:='["Group"]' format=json
HTTP/1.1 200 OK
Content-Type: application/json
Date: Mon, 20 Apr 2015 06:57:50 GMT
Server: Warp/3.0.11
Transfer-Encoding: chunked

[
    {
        "goals": [
            {
                "details": "",
                "name": "rightunit",
                "proof_tree": "",
                "prover_output": "...",
                "result": "Disproved",
                "tactic_script": {
                    "extra_options": [
                        "-DocProof"
                    ],
                    "time_limit": 1
                },
                "used_axioms": [],
                "used_prover": "SPASS",
                "used_time": {
                    "components": {
                        "hours": 0,
                        "mins": 0,
                        "secs": 0
                    },
                    "seconds": 0
                },
                "used_translation": "CASL2SoftFOL"
            },
            {
                "details": "",
                "name": "zero_plus",
                "proof_tree": "",
                "prover_output": "...",
                "result": "Proved",
                "tactic_script": {
                    "extra_options": [
                        "-DocProof"
                    ],
                    "time_limit": 1
                },
                "used_axioms": [
                    "leftunit"
                ],
                "used_prover": "SPASS",
                "used_time": {
                    "components": {
                        "hours": 0,
                        "mins": 0,
                        "secs": 0
                    },
                    "seconds": 0
                },
                "used_translation": "CASL2SoftFOL"
            }
        ],
        "node": "Group"
    }
]
```

### DIFF
--- a/Common/Utils.hs
+++ b/Common/Utils.hs
@@ -22,7 +22,7 @@ module Common.Utils
   , trim
   , trimLeft
   , trimRight
-  , toSnake
+  , toSnakeCase
   , nubOrd
   , nubOrdOn
   , atMaybe
@@ -156,20 +156,23 @@ trimRight = foldr (\ c cs -> if isSpace c && null cs then [] else c : cs) ""
    Convert CamelCased or mixedCases 'String' to a 'String' with underscores,
    the \"snake\" 'String'.
 
-   It also considers abbreviations in capital letters:
-   `toSnake "XMLHttpRequest" == "xml_http_request"`
+   It also considers SCREAMINGCamelCase:
+   `toSnakeCase "SomeSCREAMINGCamelCase" == "some_screaming_camel_case"`
 -}
-toSnake :: String -> String
-toSnake =
-  map toLower . intercalate "_" . glueSingletons . groupBy (const isLower)
+toSnakeCase :: String -> String
+toSnakeCase = toSnakeCase' False
   where
-    {- glue adjacent singleton lists together:
-       `glueSingletons "" ["abc", "d", "e", "fg" "h", "i"]
-       == ["abc", "de", "fg", "hi"]`
-    -}
-    glueSingletons :: [String] -> [String]
-    glueSingletons =
-      map concat . groupBy (\ v w -> (length v == 1) && (length w == 1))
+    toSnakeCase' :: Bool -> String -> String
+    toSnakeCase' screaming camel =
+      case camel of
+        a : r@(b : _) -> let
+          snake_tail = toSnakeCase' True r
+          snake_head = toLower a
+          in case (isLower a, isLower b) of
+            (True, False) -> a : '_' : toSnakeCase' False r
+            (False, True) | screaming -> '_' : snake_head : snake_tail
+            _ -> snake_head : snake_tail
+        _ -> map toLower camel
 
 {- | The 'nubWith' function accepts as an argument a \"stop list\" update
 function and an initial stop list. The stop list is a set of list elements
--- a/Common/Utils.hs
+++ b/Common/Utils.hs
@@ -187,20 +187,14 @@ splitR p s =
    Based on https://gist.github.com/ruthenium/3715696
 -}
 toSnake :: String -> String
-toSnake = downcase . concat . underscores . glueSingletons "" . splitR isUpper
+toSnake = downcase . intercalate "_" . glueSingletons . splitR isUpper
   where
-    underscores :: [String] -> [String]
-    underscores [] = []
-    underscores (h : t) = h : map ('_' :) t
-
-    glueSingletons :: String -> [String] -> [String]
-    glueSingletons _ [] = []
-    glueSingletons acc ([] : ws) = glueSingletons acc ws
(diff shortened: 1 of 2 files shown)


Common/Utils.hs
```diff
@@ -151,6 +154,54 @@ trimLeft = dropWhile isSpace
...
+  let
+    go :: Char -> String -> [String]
+    go m s' = case break p s' of
+      (b', []) -> [m : b']
+      (b', x : xs) -> (m : b') : go x xs
+  in case break p s of
+    (b, []) -> [b]
+    ([], h : t) -> go h t
+    (b, h : t) -> b : go h t
+
```
cmaeder: We have already splitBy (and splitByList) that look(s) shorter to me. I would prefer a better version, maybe one that only considers if an upper case letter is next to a lower case letter.

Common/Utils.hs
```diff
@@ -151,6 +154,54 @@ trimLeft = dropWhile isSpace
...
+   Based on https://gist.github.com/ruthenium/3715696
+-}
+toSnake :: String -> String
+toSnake = downcase . concat . underscores . glueSingletons "" . splitR isUpper
+  where
+    underscores :: [String] -> [String]
+    underscores [] = []
+    underscores (h : t) = h : map ('_' :) t
+
```
cmaeder: `concat . underscores` is better done using `intercalate "_"`.

Common/Utils.hs
```diff
@@ -151,6 +154,54 @@ trimLeft = dropWhile isSpace
...
+    glueSingletons :: String -> [String] -> [String]
+    glueSingletons _ [] = []
+    glueSingletons acc ([] : ws) = glueSingletons acc ws
+    glueSingletons acc ([c] : ws) = glueSingletons (acc ++ [c]) ws
+    glueSingletons acc (w : ws) =
+      if null acc
+      then w : glueSingletons "" ws
+      else acc : w : glueSingletons "" ws
+
```
cmaeder: this function is not documented and has IMHO too many cases. Maybe it can be avoided completely by a proper split function.

Common/Utils.hs
eugenk: `splitBy` and `splitByList` are not directly applicable here, because they check for equality but we need predicates. We would need a more general function, but I could not find a simple solution. Everything I can think of needs special handling with offsets per predicate (where to split when a predicate matches).

cmaeder: Yes, `splitBy` is not usable, because it removes the delimiter. `splitR` is only used with the special predicate isUpper.

It seems easier to write a single recursive function that can replace `glueSingletons "" . splitR isUpper`.

cmaeder: How should "aBCd"  be convert? "a_bCd"? And "A" to "_a"?

cmaeder: maybe `concatMap (\ c -> if isUpper c then '_' : [toLower c] else [c])` is good enough? "aBCd" goes to "a_b_cd" then.

eugenk: Strictly speaking, we want `toSnake "aBCd" == "a_b_cd"`. The reason is that a real life example `toSnake "XmlHTTPRequest" == "xml_http_request"` is desired.

eugenk: I'm afraid `concatMap (\ c -> if isUpper c then '_' : [toLower c] else [c])` is not good enough. It puts underscores between each character in the screaming part of screaming camel case:
`concatMap (\ c -> if isUpper c then '_' : [toLower c] else [c]) "XmlHTTPRequest" == "_xml_h_t_t_p_request"`

eugenk:
How about the change of 7396694?

cmaeder:
Do you need to export `splitR`? `groupBy (const isLower)` seems to do the same as `splitR isUpper`.
I'm still not happy with toSnake, though. `(intercalate "")` is `concat`.

eugenk:
I removed `splitR`, it really seems to be redundant.

eugenk:
That's amazing. I renamed some bindings for documenting reasons and allowed to call it with the string argument only.

cmaeder:
I made another version of `toSnakeCase`, but I'm not sure if it is any clearer. Existing underscores in input strings do not behave very well.

eugenk:
This is much clearer and I think for the current task we can rely on the input string being real camelCase, i.e. without underscores. Also, the documentation already states that it only converts camelCase to snake_case. So :+1: for this function.

cmaeder:
:+1:

tillmo:
:+1:

PR #1492, opened 2015-05-05
Fix proving when MathServe is not available.

eugenk:
With MathServ being offline, we get an empty string as the tactic
script. Calling `read` on it results in the error `no parse`.

To fix this, we simply return no tactic script (`Nothing`) in this case.

### DIFF
--- a/PGIP/Output/Proof.hs
+++ b/PGIP/Output/Proof.hs
@@ -83,7 +83,7 @@ formatProofs format options proofs = case format of
           else Nothing
       , usedProver = proverOrConsCheckerName proverOrConsChecker
       , usedTranslation = showComorph translation
-      , tacticScript = fmap convertTacticScript proofStatusM
+      , tacticScript = convertTacticScript proofStatusM
       , proofTree = fmap (show . LP.proofTree) proofStatusM
       , usedTime = fmap (convertTime . LP.usedTime) proofStatusM
       , usedAxioms = fmap LP.usedAxioms proofStatusM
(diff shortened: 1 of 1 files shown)


PGIP/Output/Proof.hs
```diff
@@ -108,12 +108,17 @@ formatProofs format options proofs = case format of
...
-    let atp = read $ (\ (LP.TacticScript ts) -> ts) $ LP.tacticScript ps
-    in TacticScript { timeLimit = tsTimeLimit atp
-                    , extraOptions = tsExtraOpts atp
-                    }
+  convertTacticScript :: Maybe (LP.ProofStatus G_proof_tree)
+                      -> Maybe TacticScript
+  convertTacticScript Nothing = Nothing
+  convertTacticScript (Just ps) =
+    let ts = (\ (LP.TacticScript ts) -> ts) $ LP.tacticScript ps
+        atp = read ts
```
cmaeder: maybe it is a better idea to use `Common.Uitls.readMaybe` instead of only testing the special case of an empty string.

eugenk: Great, thanks!

PGIP/Output/Proof.hs
```diff
@@ -108,12 +109,16 @@ formatProofs format options proofs = case format of
...
-    let atp = read $ (\ (LP.TacticScript ts) -> ts) $ LP.tacticScript ps
-    in TacticScript { timeLimit = tsTimeLimit atp
-                    , extraOptions = tsExtraOpts atp
-                    }
+  convertTacticScript :: Maybe (LP.ProofStatus G_proof_tree)
+                      -> Maybe TacticScript
+  convertTacticScript Nothing = Nothing
+  convertTacticScript (Just ps) =
+    let tsStr = (\ (LP.TacticScript ts) -> ts) $ LP.tacticScript ps
+    in case readMaybe tsStr of
```
cmaeder: let-in should be removed

```
case (\ (LP.TacticScript ts) -> readMaybe ts) $ LP.tacticScript ps of
```

(Also record syntax for LP.TacticScript in LP could be used to get a proper selector.)

eugenk:
> (Also record syntax for LP.TacticScript in LP could be used to get a proper selector.)

Actually I don't quite understand what's going on with the `TacticScript` in Hets in general. Why is this kept as a String in the first place? As far as I can see, it is always a String representation of `ATPTacticScript`.

I'm referring to Logic.Prover:
```haskell
data TacticScript = TacticScript String deriving (Eq, Ord, Show, Typeable)
```
and `Interfaces.GenericATPState`:
```haskell
data ATPTacticScript = ATPTacticScript
    { tsTimeLimit :: Int, -- ^ used time limit
      tsExtraOpts :: [String] -- ^ used extra options (if any)
    } deriving (Eq, Ord, Show, Read)
```

cmaeder:
The `LP.TacticScript` is more general (also for interactive theorem provers) and is therefore just a wrapped `String`. 

Automatic theorem provers (ATPs) are supposed to just be controlled by some (command line) options and a time limit.

Maybe the distinction is unnecessary and could be removed, but I do not remember when it was introduced (and by whom).

:+1: for the current change

eugenk:
Thanks for the clarification and the code suggestions.

PR #1478, opened 2015-03-26
1477 REST: add display name to the provers list

eugenk:
This shall fix #1477. The displayName is not as beautiful as in the issue text, but as far as I can see, this is the best we have.

This branch is based on the branch of #1476 (**1475-use_consistent_naming_of_provers**) instead of **master**.  It might be easier to review that one first. (I needed the `PGIP.Common` module here as well.)

For
```
curl http://localhost:8000/provers/http%3A%2F%2Fdevelop.ontohub.org%2Fdefault%2Fpartial_order%2F%2Fstrict_partial_order/auto\?format\=json
```
on an Ubuntu machine I get
```json
{
"provers": [{
 "name": "QuickCheck", "displayName": "QuickCheck" }, {
 "name": "isabelle_process", "displayName": "isabelle_process" }, {
 "name": "Isabelle-jedit", "displayName": "Isabelle-jedit" }, {
 "name": "Isabelle-emacs", "displayName": "Isabelle-emacs" }, {
 "name": "eprover", "displayName": "eprover" }, {
 "name": "darwin", "displayName": "darwin" }, {
 "name": "darwin-non-fd", "displayName": "darwin-non-fd" }, {
 "name": "Vampire", "displayName": "Vampire" }, {
 "name": "MathServeBroker", "displayName": "MathServe Broker" }, {
 "name": "SPASS", "displayName": "SPASS" }, {
 "name": "Isabellesledgehammer",
 "displayName": "Isabelle (sledgehammer)"
 }, {
 "name": "Isabellerefute", "displayName": "Isabelle (refute)" }, {
 "name": "Isabellenitpick", "displayName": "Isabelle (nitpick)" }, {
 "name": "Isabelleautomated", "displayName": "Isabelle (automated)"
 }]
}
```

PR #1493, opened 2015-05-05
Add display name to prove-output.

eugenk:
In Ontohub, we need the `used_prover` key of the `prove` command to include the `display_name` as well. I didn't think about it when I created #1478.

### DIFF
--- a/PGIP/Common.hs
+++ b/PGIP/Common.hs
@@ -1,7 +1,0 @@
-module PGIP.Common where
-
-import Proofs.AbstractState
-
-data ProverOrConsChecker = Prover G_prover
-                         | ConsChecker G_cons_checker
-                         deriving (Show)
--- a/PGIP/Output/Formatting.hs
+++ b/PGIP/Output/Formatting.hs
@@ -9,9 +9,6 @@ import Proofs.AbstractState
 
 import Data.Char
 
-proverOrConsCheckerName :: ProverOrConsChecker -> String
-proverOrConsCheckerName = mkNiceProverName . internalProverName
-
 internalProverName :: ProverOrConsChecker -> String
 internalProverName pOrCc = case pOrCc of
   Proofs.AbstractState.Prover pr -> getProverName pr
@@ -29,7 +26,7 @@ getWebProverName :: G_prover -> String
 getWebProverName = mkNiceProverName . getProverName
 
 proversOnly :: [(AnyComorphism, [ProverOrConsChecker])] -> [ProverOrConsChecker]
-proversOnly = nubOrdOn proverOrConsCheckerName . concatMap snd
+proversOnly = nubOrdOn (mkNiceProverName . internalProverName) . concatMap snd
 
 showProversOnly :: [(AnyComorphism, [String])] -> [String]
 showProversOnly = nubOrd . concatMap snd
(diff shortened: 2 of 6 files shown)


cmaeder:
Could the difference between `proverOrConsCheckerName` and `internalProverName` just be computed by `removeFunnyChars`?

eugenk:
Sorry, I don't understand the question. `removeFunnyChars` is used in `proverOrConsCheckerName`.

cmaeder:
Is it enough to only store `internalProveName` and compute `proverOrConsCheckerName` by applying `removeFunnyChars`? (I'm away now, sorry)

eugenk:
I actually don't want to change the use of `proverOrConsCheckerName` because it is used as an identifier for the prover. `internalProverName` on the other hand just gets the name that is in the record.

eugenk:
The `name` must be exactly what Hets recognizes as a prover on a `prove` request. I don't quite understand the mechanics of finding the prover by a string in the RESTful server, so I'd rather not change this.

cmaeder:
The answer to my above question is yes, therefore `internalProverName` should be moved to `PGIP.Output.Formatting`! 

```
proverOrConsCheckerName = removeFunnyChars . internalProverName
```

The field `name` in PGIP.Output.Provers.Prover is not really needed.
Therefore the current `name` could be `removeFunnyChars . displayName`.

it is confusing that the internal name is denoted as display name, so maybe `displayName` could be renamed. (Also `name` might be removed or renamed as it is a too generic identifier and difficult to grep for).

Unfortunately, `getWebProverName` is used (once only) in `PGIP.Server.filterByProver`, so it should stay as given in `PGIP.Output.Formatting`

eugenk:
> The field name in PGIP.Output.Provers.Prover is not really needed.

It most certainly is, because we need it in the output.

cmaeder:
`name` does not need to be a (redundant) record field! it could be defined via:

```
name = removeFunnyChars . displayName
```

eugenk:
The point is that the data type is serialized automatically and - as far as I understand it - we need a record field for this serialization.

eugenk:
Note that the point of the data type is only to provide easy serializiation (without the need to write a function for each xml-tag/json-entry).

cmaeder:
I see, for automatic serialization `name` should remain a field. Yet, `internalProverName` should be moved to `Formatting`. (`proverOrConsCheckerName` is also used only once in `Server`.)

To express the redundancy within `Provers.Prover` a custom constructor should be defined and used in `prepareFormatProver` (or elsewhere):

```
mkProver s = Prover { displayName = s, name = removeFunnyChars s }
```

The actual Constructor `Prover` should not be exported (unless it is needed for serialization).

Side remark: maybe  the module `PGIP.Common` can be removed and  `ProverOrConsChecker` be defined in `Proofs.AbstractState`.

cmaeder:
Maybe, `removeFunnyChars` should be renamed to something like `mkNiceProverName` and used elsewhere (manually inlining `proverOrConsCheckerName` and `getWebProverName` in `Server`).

cmaeder:
Maybe also more qualified names can be avoided (as I just saw in `Server` that moving the data type caused additional changes.)

eugenk:
I think I have implemented all your changing requests.

eugenk:
Oh, I missed `getWebProverName`, sorry.

cmaeder:
it's fine from my side now :+1:

eugenk:
Alright. Then I'm only waiting for travis.

Issue #1510, opened 2015-08-11
Analysing a TPTP file doesn't yield symbols


eugenk:
If, for example, I analyse a Casl file, the xml output contains a `<Symbols>` section. For a TPTP file, the output doesn't have that section. This needs to be fixed for ontohub's axiom selection to work. Also, I need this feature for the evaluation of my master's thesis.

eugenk:
Closed with #1511

PR #1511, opened 2015-08-12
extended static analysis

cmaeder:
@eugenk for #1510 static analysis for the logic `SoftFOL` needed to be implemented


eugenk:
:+1:
Many thanks for implementing this :)

PR #1517, opened 2015-08-24
Add more clif-specifics to magic file.

### LINKED ISSUE #1516: Hets filetype detection broken for clif
Hets cannot detect the clif file type of (http://ontohub.org/ref/2/imageschemafamily/path/movementAlongPath)

```
curl http://localhost:8000/dg/http%3A%2F%2Fontohub.org%2Fref%2F2%2Fimageschemafamily%2Fpath%2FmovementAlongPath/full-signatures/full-theories/auto
```

while it can detect it with the url (https://ontohub.org/repositories/imageschemafamily/a9b7796881d3e70a677a846aebd56925841d5914/download/path/movementAlongPath.clif)
```
curl http://localhost:8000/dg/https%3A%2F%2Fontohub.org%2Frepositories%2Fimageschemafamily%2Fa9b7796881d3e70a677a846aebd56925841d5914%2Fdownload%2Fpath%2FmovementAlongPath.clif/full-theories/full-signatures
```

Note that both URLs point to the exact same file. We assume that the latter works because it has a file extension while the former has not.

eugenk:
This should fix #1516. Using `file -m ` with the new magic file, the referenced file from the issue is recognised as CLIF.

### DIFF
--- a/magic/hets.magic
+++ b/magic/hets.magic
@@ -124,7 +124,7 @@
 0       search/4096/t   \(or\                   CLIF document
 !:mime  text/clif
 
-0       search/4096/t   \if(\                   CLIF document
+0       search/4096/t   \(if\                   CLIF document
 !:mime  text/clif
 
 0       search/4096/t   \(iff\                  CLIF document
--- a/magic/hets.magic
+++ b/magic/hets.magic
@@ -94,6 +94,12 @@
 0       search/4096/t   \(cl-text\              CLIF document
 !:mime  text/clif
 
+0       search/4096/t   \(cl-module\            CLIF document
+!:mime  text/clif
+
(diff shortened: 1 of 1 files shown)


magic/hets.magic
```diff
@@ -103,6 +109,30 @@
...
+0       search/4096/t   \(exists\               CLIF document
+!:mime  text/clif
+
+0       search/4096/t   \(and\                  CLIF document
+!:mime  text/clif
+
+0       search/4096/t   \(or\                   CLIF document
+!:mime  text/clif
+
+0       search/4096/t   \if(\                   CLIF document
```
tillmo: should be "(if", not "if("

tillmo:
:+1:

PR #1521, opened 2015-09-28
Lower priority of clif

eugenk:


cmaeder:
If moving the entries helps then :+1:

PR #1530, opened 2015-10-20
changed analysis of string such that type of imported files in libraries can be guessed

mcodescu:
If a string is a file name, we should be able to guess its type and parse it with the appropriate tool.

### DIFF
--- a/Static/AnalysisLibrary.hs
+++ b/Static/AnalysisLibrary.hs
@@ -133,8 +133,10 @@ anaString mln lgraph opts topLns libenv initDG input file mr = do
       posFileName = case mln of
           Just gLn | useLibPos opts -> libToFileName gLn
           _ -> if checkUri file then file else realFileName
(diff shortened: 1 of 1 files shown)


cmaeder:
```
Static/AnalysisLibrary.hs:136:57: trailing (1) white space
Static/AnalysisLibrary.hs:137:27: put blank before {
Static/AnalysisLibrary.hs:139:14: put blank after <-
```

mcodescu:
Christian, thanks for your change! Does this mean that we can now merge this pull request?

cmaeder:
This change simply ignores the `-i` setting of the command line. `readLibDefn` is called for the top-level library and imported libraries. 

I don't know how to check for the top-level or imports (maybe by looking into `LNS`).

cmaeder:
forget this PR and use https://github.com/spechub/Hets/pull/1534

PR #1532, opened 2015-10-21
insert String encoding for utf-8

### LINKED ISSUE #1531: DGXML contains non-UTF8 characters
e.g. for http://ontohub.org/meta/dol-ontology.owl, Hets produces an XML that contains 0xE9 characters. Try
```
curl localhost:8000/dg/file%3a%2f%2f%2f^Ctohub.org%2fmeta%2fdol-ontology/full-signatures/full-theories/auto |hd |grep " e9"
```
Actually, this problem only occurs with the hets server, not with hets command line.

cmaeder:
fixes #1531

### DIFF
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -26,6 +26,8 @@ import Driver.Version
 
 import Network.Wai.Handler.Warp
 import Network.HTTP.Types
+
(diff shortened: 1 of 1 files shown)


tillmo:
:+1:

PR #1533, opened 2015-10-21
added missing dependency

cmaeder:
for calling `cabal build` in the top-level-directory

PR #1537, opened 2015-10-28
Fix compilation issues

gnn:
Using ghc 7.10.1 I had compilation issues because:

  - `Data.CaseInsensitive` is a member of a package which is hidden by
    default and needs to be added to `build-depends` in order to be
    unhidden and
  - Haskell's `if` needs its `else` to be indented by one more level.

### DIFF
--- a/Hets.cabal
+++ b/Hets.cabal
@@ -71,6 +71,7 @@ Executable hets
     , array >= 0.1.0.0
     , aterm >= 0.1.0.1
     , base >= 4.0.0.0
+    , case-insensitive
     , containers >= 0.1.0.1
     , directory >= 1.0.0.0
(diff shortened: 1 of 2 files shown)


jelmd:
FWIW: found the deps in the http-types, wai-extra, wai-logger and warp module

cmaeder:
please review and merge my PR #1533 first

cmaeder:
Only review then I'll merge

cmaeder:
You can also merge this PR and close mine :+1:

cmaeder:
It is also possible to let travis check the compilation with various versions of ghc.

cmaeder:
this PR is now obsolete after #1533

PR #1539, opened 2015-11-03
fix #1538

### LINKED ISSUE #1538: clarify use of the "empty.txt"
In `PGIP.Server`, we have two places where the string `"empty.txt"` is used. This seems to be a filename for synchronisation.

Why exactly is this file there? Why is it called "empty.txt"? It seems to restrict the Hets server to one running instance per machine.

A better filename could be something like "hets-server-instance-[port-number].lock", if it is used for synchronisation.

cmaeder:


cmaeder:
The port may be given via the command line `hets -X -S 8800`. Regarding the name of the log file I don't care. (Maybe it should be command line option, too.) Creating a random temp file (same price) may make it difficult to find the log file. What needs to be safer? I prefer higher-level libraries over lower-level ones (like `System.Posix`)

eugenk:
:+1:

eugenk:
Oh wait, you could always print https://github.com/spechub/Hets/pull/1539/files#diff-3c5ab7ca1adb7f19e37703e26e6b050aR192 and not just in verbose mode.

cmaeder:
Usually, only `-v2` shows file names read or written to.

eugenk:
Okay, then :+1:

jelmd:
With "safer" I meant if there are several instances running on the same machine, getting into trouble is very easy as long as such "ought to be writable directories" are unintentionally "shared". E.g. in our labs there runs at least one hets instance as user 'webservd' or whatever and assumes that it is the owner of e.g. /tmp/MyHetsLib - true for the first instance which gets fired. However, if a stud/other user wants to play around with hets, changing the port may not help, because he simply isn't allowed to write to /tmp/MyHetsLib/...

However, if an hets instance would do the equiv of "mkdir -t -d MyHetsLib.XXXXXX", haskell/OS guarantees, it would get a new, uniq and therefore its very own directory -> no trouble anymore.

And if each instance has its own directory, it does not need to think about "haring" things or invent new wheels for making things uniq/prevent sharing, e.g. the port in the logfile name would be redundant as well.

If it is required to be able to map a directory/file to an instance, this shouldn't be a problem as well, because the tempfile/dir prefix (the part infront of the 6X) can be used in the same manner it is done right now for the logfile, however, IMHO the PID ($$) would always be much much better, than an "arbitrary" port.

PR #1540, opened 2015-11-03
Partially solve 1494 parse dol standard examples

gnn:
The commits on this branch solve some of the parsing issues hets has with the DOL standard.
Two outstanding issues still remain:
- issues with language `serialization` CURIES (so don't use those yet when you're testing)
- `select` and `reject` can not handle `SymbolList`s

which is why the ticket can not be fully closed yet. The fixes for those issues will follow soon.
I'm creating this pull request before fully solving this issue because one can still use the parser on quite a few things, so it might be of some help already.

I tested for conflicts. There shouldn't be any.

### DIFF
--- a/Common/AnnoParser.hs
+++ b/Common/AnnoParser.hs
@@ -27,6 +27,7 @@ module Common.AnnoParser
     , newlineOrEof
     ) where
 
+import Text.Parsec.Char (endOfLine)
 import Text.ParserCombinators.Parsec
 import Text.ParserCombinators.Parsec.Error
 import Text.ParserCombinators.Parsec.Pos as Pos
@@ -237,7 +238,8 @@ floatingAnno ps = literal2idsAnno ps Float_anno
 
 prefixAnno :: Range -> GenParser Char st Annotation
 prefixAnno ps = do
-    prefixes <- many $ do
+    prefixes <- many $ skipMany (commentLine >> endOfLine) >> do
+        spaces
         p <- (string colonS >> return "") <|>
              (IRI.ncname << string colonS)
         spaces
--- a/Common/AnnoParser.hs
+++ b/Common/AnnoParser.hs
@@ -27,7 +27,6 @@ module Common.AnnoParser
     , newlineOrEof
     ) where
 
-import Text.Parsec.Char (endOfLine)
 import Text.ParserCombinators.Parsec
 import Text.ParserCombinators.Parsec.Error
 import Text.ParserCombinators.Parsec.Pos as Pos
@@ -238,8 +237,7 @@ floatingAnno ps = literal2idsAnno ps Float_anno
 
 prefixAnno :: Range -> GenParser Char st Annotation
 prefixAnno ps = do
-    prefixes <- many $ skipMany (commentLine >> endOfLine) >> do
-        spaces
+    prefixes <- many $ skipMany (commentLine >> spaces) >> do
         p <- (string colonS >> return "") <|>
              (IRI.ncname << string colonS)
         spaces
--- a/Common/Keywords.hs
+++ b/Common/Keywords.hs
@@ -309,6 +309,9 @@ interpretationS = "interpretation"
 moduleS :: String
 moduleS = "module"
 
+omsS :: String
+omsS = "oms"
+
 ontologyS :: String
 ontologyS = "ontology"
 
--- a/Common/Keywords.hs
+++ b/Common/Keywords.hs
@@ -507,6 +507,9 @@ intersectS = "intersect"
 lambdaS :: String
 lambdaS = "lambda"
 
+languageS :: String
+languageS = "language"
+
 left_assocS :: String
 left_assocS = "left_assoc"
 
--- a/Common/Token.hs
+++ b/Common/Token.hs
@@ -119,7 +119,7 @@ terminatingKeywords =
 startingKeywords :: [String]
 startingKeywords =
     [ archS, fromS, logicS, newlogicS, refinementS, specS, unitS, viewS
-    , ontologyS, alignmentS, networkS, equivalenceS, newcomorphismS
+    , omsS, ontologyS, alignmentS, networkS, equivalenceS, newcomorphismS
     , interpretationS, entailmentS ]
 
 -- | keywords that may follow a defining equal sign
--- a/Common/Token.hs
+++ b/Common/Token.hs
@@ -112,7 +112,7 @@ criticalKeywords = terminatingKeywords ++ startingKeywords
 -- | keywords terminating a basic spec
 terminatingKeywords :: [String]
 terminatingKeywords =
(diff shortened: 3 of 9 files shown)


gnn:
Please don't merge the pull request yet. There a are a few more commits to come.

cmaeder:
The policy is that (only) the creator of a PR merges after an OK by another reviewer

Common/AnnoParser.hs
```diff
@@ -237,7 +238,8 @@ floatingAnno ps = literal2idsAnno ps Float_anno
 
 prefixAnno :: Range -> GenParser Char st Annotation
 prefixAnno ps = do
-    prefixes <- many $ do
+    prefixes <- many $ skipMany (commentLine >> endOfLine) >> do
```
cmaeder: I'm not sure if you really need `endOfLine`. I suspect that `spaces` also skips `crlf`. If you really need to check for the end of a line, then `newline` (from `Text.ParserCombinators.Parsec.Char`) should do, too.

cmaeder: Furthermore, comment lines should not be skipped but kept as part of the global annotations

tillmo:
The problem that select and reject can not handle SymbolLists is solved by branch `filtering_with_symbol_lists`

mcodescu:
The change suggested by Christian (newline instead of endOfLine) is enough to make the compilation go through (and probably the tests green). I have tried the files from https://ontohub.org/dol-examples. There are some that can't go through because of missing translations or logics. For some we would need more of the static analysis for DOL that has been implemented in the meanwhile in HETS (probably via a rebase). The only problem I could find so far is for https://ontohub.org/dol-examples/sorting.dol
```
./hets ~/repos/dol-examples/sorting.dol 
### file name '/home/mcodescu/repos/dol-examples/sorting.dol' does not match library name 'Sorting'
Analyzing library Sorting
logic CASL
Analyzing spec Sorting?TotalOrder
Analyzing spec Sorting?Nat
Analyzing spec Sorting?List
Analyzing spec Sorting?Sorting
/home/mcodescu/repos/dol-examples/sorting.dol:48.1-48.4: *** Error:
attempt to hide the following symbols from the local environment:
{pred is_ordered : List, pred permutation : List * List}
hets: user error (Stopped due to errors)

```
Hiding from a named specification works though. This seems to be a parsing error.

gnn:
I tested this and it seems that the problem is the second [`then`](https://gist.github.com/gnn/10be7711f53d9bca2d4416d2ba54fc32/926f6dc871b1d6b8b4d8be12785dcbf816bbb732#file-sorting-dol-L11) of the `spec Sorting`. I think it creates a nested scope and the `hide is_ordered, permutation` only affects that scope, which doesn't work, as the symbols it attempts to hide also appear outside of that scope. Adding curly braces, [see this diff](https://gist.github.com/gnn/10be7711f53d9bca2d4416d2ba54fc32/revisions?diff=split), to explicitly scope the `hide` over the complete `spec` squelches the error. As a second option, removing the [`then`](https://gist.github.com/gnn/10be7711f53d9bca2d4416d2ba54fc32/926f6dc871b1d6b8b4d8be12785dcbf816bbb732#file-sorting-dol-L11) has the same effect.
I don't know whether this behaviour of `hide` is intended though, so I'd like to ask whether I should change the example according to the linked [gist](https://gist.github.com/gnn/10be7711f53d9bca2d4416d2ba54fc32) or modify how `hide` behaves.

Common/AnnoParser.hs
gnn: Thanks for suggesting `spaces`. It worked and also enabled me to get rid of a line that I wanted to be gone from the get go. See 868229b below.
The reason I'm skipping the comments is that `prefixAnno` is already parsing a `%prefix( ... )%` annotation block and it needs to return a [`Prefix_anno`](https://github.com/spechub/Hets/blob/b259e3b3e05433b2018f45d5781000fa3af7cbdc/Common/AS_Annotation.der.hs#L108). Since I wasn't able to figure out where to put the `Unparsed_anno` returned from `commentLine` I decided to skip those comments.

tillmo:
@mcodescu, could you please have a look at the last two commits? Can we at least merge this pull request now, even if not all problems have been solved? (The pull request should of course not break existing functionality.)

mcodescu:
I have parsed again the specifications in the repository mentioned above. Now the sorting specification is parsed correctly. It might be a good idea to run more tests before merging, though, but I don't have more inputs. Also, when compiling this branch on my machine, I had to take care of some ambiguous occurrences of methods, especially involving Data.Graph.Inductive.Graph, which is usually not imported using qualified. So if you do merge, please make sure that you don't introduce compilation errors.

tillmo:
please rebase this branch on master (`git rebase master`, solve problems, then `git push -f`) in order to see if it works with the recently added tests.

gnn:
I rebased the branch onto the current master. There are no compilation errors and I tried parsing all the DOL examples in the repository. There are still quite a few errors, most of which stem from missing logics or comorphisms, but some also have other reasons.

These are the categories of parsing errors, encountered when trying `hets -p` on the contents of the current `dol-examples` repository, missing `.env` files excluded and in order of decreasing frequency.
The errors are links to gists containing the error messages:

#### [Missing comorphisms](https://gist.github.com/gnn/0988e5c2c32e1e59094b310f54c1eb04#file-missing-comorphisms)

Pretty simple. Hets doesn't seem to know the comorphisms uses in these examples, so the parser complains.

#### [Unknown logics](https://gist.github.com/gnn/0988e5c2c32e1e59094b310f54c1eb04#file-unknown-logics)

Similar to the above. Hets doesn't seem to know one of the logics used in the examples. Note that the `purl.net` prefix is not the culprit in the first instance. Hets doesn't know the logic, even if this prefix is removed. Also beware: the second example is mentioned here only because this is the first type of error encountered. There probably is a missing comorphism in this example too.

#### [Unknown keyword `distributed-ontology`](https://gist.github.com/gnn/0988e5c2c32e1e59094b310f54c1eb04#file-distributed-ontology)

It seems, that `hets` either can't handle the keyword `distributed-ontology` yet, or it is not used correctly in the example.

#### [Errors caused by unicode symbols which `hets` doesn't to be able to handle](https://gist.github.com/gnn/0988e5c2c32e1e59094b310f54c1eb04#file-unicode-symbol-problems)

I actually have a fix for this replacing the unicode symbols by their ASCII counterparts prepared locally. I just have to push the proposal to the examples repository.

#### [Unknown cause](https://gist.github.com/gnn/0988e5c2c32e1e59094b310f54c1eb04#file-unknown-cause)

This is only one file and the error is caused by the `select x where...` construct. Even if I change the construct to `select x . where...` there are errors after that one too. So I don't really know how the `select x where...` construct should be used.

gnn:
Closed by accident when commenting. Sorry.

tillmo:
`distributed-ontology` should be replaced by `library`. Where does the `select` problem occur exactly?

PR #1542, opened 2015-11-05
1541 hets fails after a many rest calls in short time

### LINKED ISSUE #1541: Hets fails after a many REST calls in short time
In Ontohub, we make a lot of calls in short time and Hets fails with `nextSess` after a while. It seems like there are too many open sessions.

What are the `Session` and the function `nextSess` for?

eugenk:
This shall fix #1541.

### DIFF
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -731,8 +731,8 @@ addNewSess sessRef sess = do
   let s = sess { sessKey = k }
   addSess sessRef s
 
-nextSess :: Session -> Cache -> LibEnv -> Int -> IO Session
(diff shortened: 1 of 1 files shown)


cmaeder:
I think `nextSess` can be called several times by continuous prove calls on the same DG. So making as session cleanable may still lead to the error. Could you try to provoke this error?

Otherwise the patch looks fine :+1:

eugenk:
I'm currently trying to provoke the error the same way I found it in the first place. The problem still persists.

eugenk:
Increasing the session map size helped a little but it is not a solution.

PR #1550, opened 2015-11-27
Fix reading long request bodies.

eugenk:
Hets only parsed the first chunk of the request body. This pull request makes it parse the whole body.


cmaeder:
good. I was not aware that we had a problem. :+1:

PR #1552, opened 2015-12-08
1506 glade gtk13

cmaeder:
this patch can be merged to support GTK with ghc-7.10 in the future.
For ghc-7.10 a new glade package still needs to be uploaded to hackage.
Until then https://github.com/cmaeder/glade can be used. 

The branch https://github.com/spechub/Hets/tree/travis_ghc-7.10 shows that it works for ghc-7.10.2.


cmaeder:
Could someone give me a :+1: (or should I proceed without?). I don't want to wait until other changes introduce merge conflicts (like for #1537)

PR #1577, opened 2016-03-03
Medusa

mcodescu:


OWL2/Medusa.hs
```diff
@@ -0,0 +1,131 @@
...
+
+--  for each individual "p1" that has a fact "meets p2"
+--  look for individuals "i1" and "i2" such that
+--  i1 has_fiat_boundary p1 and i2 has_fiat_boundary p2
+--  and return i1 type(p1) i2 type(p2)
+getMeetsFacts :: [Axiom] -> Set.Set (QName, QName) -> QName ->
+              Set.Set (QName, QName, QName, QName)
+getMeetsFacts axs tInds n = case mapMaybe (getMeetsFactsAux axs tInds n) axs of
+   [] -> Set.empty
+   x -> Set.fromList x
```
tillmo: you can avoid the case distinction and use `Set.fromList` in both cases.

mcodescu: Right, corrected, see commit.

OWL2/Medusa.hs
```diff
@@ -0,0 +1,131 @@
...
+
+getFiatBoundaryFacts :: [Axiom] -> Set.Set (QName, QName) -> QName -> QName ->
+                     Maybe (QName, QName, QName, QName)
+getFiatBoundaryFacts axs tInds point1 point2 =
+   let i1 = case mapMaybe (getFiatBoundaryFactsAux point1) axs of
+              (c:_) -> Just c
+              [] -> Nothing
+       i2 = case mapMaybe (getFiatBoundaryFactsAux point2) axs of
+              (c:_) -> Just c
+              [] -> Nothing
```
tillmo: this pattern occurs 3 times (also in function `getClass`, if you add a call of function `maybe` there) and should be relegated to an own function.

mcodescu: Of course. Done

tillmo:
I can restore the java files (see http://stackoverflow.com/questions/1186535/how-to-modify-a-specified-commit-in-git how to do this), and would like to force-push the result (then the files appear to have never been deleted). Please tell me once you are ready with your changes, because a force-push badly interferes with other changes.

eugenk:
Just as a general comment (not something you would need to change in this pull request):

Instead of creating a method `medusaToJson` that directly creates a JSON data structure, you can also create a method `buildMedusa` that creates an ordinary Haskell data structure that can be serialised to JSON by `ppJson . toJson`, like it's done in, for instance, https://github.com/spechub/Hets/blob/master/PGIP/Output/Provers.hs#L51-L59 (and the method `formatAsJson`/`formatAsXml` in this file)

I find this a little better maintainable and it also allows to serialise to other formats easily (currently only XML is implemented; not really important for medusa).

tillmo:
But we are doing exactly this!

tillmo:
OK, I have restored the java files.

tillmo:
:+1:

PR #1580, opened 2016-03-03
1572 static analysis for intersections

mcodescu:
I did not implement the analysis for heterogeneous intersections. But this can be done later.


Logic/ExtSign.hs
```diff
@@ -65,6 +66,17 @@ ext_signature_union l e1@(ExtSign s1 _) e2@(ExtSign s2 _) = do
...
     return $ makeExtSign l s
 
+ext_signature_intersect :: Logic lid sublogics
+        basic_spec sentence symb_items symb_map_items
+        sign morphism symbol raw_symbol proof_tree
+        => lid -> ExtSign sign symbol -> ExtSign sign symbol
+               -> Result (ExtSign sign symbol)
+ext_signature_intersect l e1@(ExtSign s1 _) e2@(ExtSign s2 _) = do
+    checkExtSign l "u1" e1
+    checkExtSign l "u2" e2
```
tillmo: please use some more informative messages instead of "u1" and "u2". E.g. "left-hand side of intersection" and "right-hand side of intersection".

Logic/Prover.hs
```diff
@@ -88,6 +89,13 @@ joinSensAux s1 s2 = let
 joinSens :: (Ord a, Eq b) => ThSens a b -> ThSens a b -> ThSens a b
 joinSens s1 = fst . joinSensAux s1
 
+intersectSens :: (Ord a, Eq b) => ThSens a b -> ThSens a b -> ThSens a b
+intersectSens s1 s2 = let
+  sens1 = map sentence $ OMap.elems s1
+  sens2 = map sentence $ OMap.elems s2
+  sensI = Set.intersection (Set.fromList sens1) (Set.fromList sens2)
+ in toThSens $ map (AS_Anno.makeNamed "") $ Set.toList sensI
```
tillmo: shouldn't the annotations of the original sentence be kept here? Perhaps just unite them.

Static/AnalysisStructured.hs
```diff
@@ -686,6 +693,45 @@ anaUnion addSyms lg ln dg nsig name opts eo asps rg = case asps of
...
+anaIntersect addSyms lg libEnv ln dg nsig name opts eo asps rg = case asps of
+  [] -> fail "empty intersection"
+  _ -> do
+      let sps = map item asps
+      (sps', nsigs, dg', _) <-
+          let ana (sps1, nsigs, dg', n) sp' = do
+                let n1 = inc n
+                (sp1, nsig', dg1) <-
+                  anaSpec addSyms lg libEnv ln dg' nsig n1 opts eo sp' $ getRange sp'
+                return (sp1 : sps1, nsig' : nsigs, dg1, n1)
```
tillmo: function `ana` should be moved to before the binding `(sps', nsigs, dg', _) <-`, at the same level as `let sps = map item asps`. This will increase readability.

Logic/Prover.hs
mcodescu: Aren't these the names of the sentences? Then we can't keep them in the intersection and can't
unite them.

tillmo: For each sentence appearing in the intersection, we have `n` (possibly different) names of the sentence in the `n` OMS to be intersected. We could append these `n` names to a new name, and take a "union" insofar as duplications in the list of `n` names are removed first.

tillmo:
turned this into an issue, see #1589

tillmo:
:+1:

PR #1592, opened 2016-03-11
1564 improve analysis of alignments

mcodescu:


tillmo:
Previously,
```
logic OWL

ontology Woman = Class: Woman 
ontology Person = Class: Person
alignment Alignment1 : Woman to Person =
  Woman < Person
end

ontology AlignedOntology1 =
  combine Alignment1
end
```
in the colimit lead to
```
Class: Person
Class: Woman
       SubClassOf: Person
```
Now we get
```
Class: Person_target
Class: Woman_source
       SubClassOf: Person_target
```
I recall that you have told me why this is so, but still I want to discuss if there is a way to get the old version, which looks nicer.

mcodescu:
I think we would need to test whether a symbol name that appears on the left side in a correspondence also appears on the right side. If this is the case, we must disambiguate. If that is not the case, we can keep the original names.

tillmo:
good idea

tillmo:
remaning issue relegated to #1597, hence :+1:

PR #1593, opened 2016-03-11
Skolemization and prenex normal form

mcodescu:


tillmo:
The comorphism `CASL2Prenex` does more than translation to prenex form: it also translates the matrix of the formula. It hence seems that `CASL2PrenexKNF` would be a better name, although I would prefer to separate this into two comorphisms. Moreover, the KNF translation does not work properly. For example, in `Basic/RelationsAndOrders.casl`, spec `RichTotalOrder`,
```
forall x, y : Elem . x < y <=> x <= y /\ not x = y
                                      %(less_def_ExtPartialOrder)%
```
gets translated to
```
forall x, y : Elem
. (x < y \/ (x = y \/ not x <= y))
  /\ ((not x = y /\ x <= y) \/ not x < y)                  %(Ax2)%
```
but this is not in KNF.

Aha, and please keep axiom labels.

Moreover,
```
forall x, y, z : Elem
. sup(x, y) = z
  <=> x <= z /\ y <= z
      /\ forall t : Elem . x <= t /\ y <= t => z <= t
                                       %(sup_def_ExtPartialOrder)%
```
gets translated to
```
forall x, y, z : Elem
. (sup(x, y) = z
   \/ ((exists t : Elem . not z <= t /\ (y <= t /\ x <= t))
       \/ not y <= z \/ not x <= z))
  /\ (((forall t : Elem . z <= t \/ (not y <= t \/ not x <= t))
       /\ y <= z /\ x <= z)
      \/ not sup(x, y) = z)                                %(Ax4)%
```
but this is not in prenex form.

mcodescu:
One more issue: I think for CASL2Skolem we must assume that signature morphisms are always identities, otherwise we can't always map the Skolem functions.

mcodescu:
My tests go all through now, see some in https://ontohub.org/sandbox/normalForms.dol

Comorphisms/CASL2NNF.hs
```diff
@@ -0,0 +1,100 @@
...
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2NNF = CASL
+    sourceSublogic CASL2NNF = SL.caslTop
+    targetLogic CASL2NNF = CASL
+    mapSublogic CASL2NNF  = Just -- TODO: does the sublogic change?
+    map_theory CASL2NNF = mapTheory
+    map_morphism CASL2NNF = return -- morphisms are mapped identically
+    map_sentence CASL2NNF _ s = return $ negationNormalForm s
+    map_symbol CASL2NNF _ = Set.singleton . id
+    has_model_expansion CASL2NNF = True -- check
```
tillmo: this clearly is true, since the model functor is the identity (this argument could be added as a comment).

Comorphisms/CASL2NNF.hs
```diff
@@ -0,0 +1,100 @@
...
+    is_weakly_amalgamable CASL2NNF = True --check
```
tillmo: dto. here

Comorphisms/CASL2NNF.hs
```diff
@@ -0,0 +1,100 @@
...
+               Symbol RawSymbol ProofTree
+               CASL CASL_Sublogics
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2NNF = CASL
+    sourceSublogic CASL2NNF = SL.caslTop
+    targetLogic CASL2NNF = CASL
+    mapSublogic CASL2NNF  = Just -- TODO: does the sublogic change?
```
tillmo: The sublogic can change from `Horn` to `FOL`, as the following example shows:
```
spec Horn =
  sort s
  preds p,q:s
  forall x:s . p(x) => q(x)
end
```
Otherwise, I think the sublogic cannot change.

Comorphisms/CASL2NNF.hs
```diff
@@ -0,0 +1,100 @@
...
+   Junction Con
+     (map (\x -> negationNormalForm $ Negation x nullRange) sens)
+   nullRange
+ Negation (Quantification Unique_existential _vars _sen _) _->
+    error "negation normal form for unique existentials nyi"
+ Negation (Quantification q vars qsen _) _ ->
+   Quantification (dualQuant q) vars
+     (negationNormalForm $ Negation qsen nullRange)
+     nullRange
+ x -> x
```
tillmo: this catchall case leaves negations of implications and of equivalences alone. Consider e.g.
```
spec Negation =
  sort s
  preds p,q:s
  forall x:s . not (p(x) => q(x))
  forall x:s . not (p(x) <=> q(x))
end
```

Comorphisms/CASL2NNF.hs
```diff
@@ -0,0 +1,100 @@
...
+ Negation (Negation sen' _) _ ->
+   negationNormalForm sen'
+ Negation (Junction Con sens _) _ ->
+   Junction Dis
+     (map (\x -> negationNormalForm $ Negation x nullRange) sens)
+     nullRange
+ Negation (Junction Dis sens _) _ ->
+   Junction Con
+     (map (\x -> negationNormalForm $ Negation x nullRange) sens)
+   nullRange
```
tillmo: these two cases could be combined into one, using a function `dualJunct :: Junctor -> Junctor`.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree
+               CASL CASL_Sublogics
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2Prenex = CASL
+    sourceSublogic CASL2Prenex = SL.caslTop --TODO
```
tillmo: why TODO here? I think `caslTop`is correct.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+    targetLogic CASL2Prenex = CASL
+    mapSublogic CASL2Prenex sl = Just sl -- TODO
+    map_theory CASL2Prenex = mapTheory
+    map_morphism CASL2Prenex = error "nyi"
```
tillmo: this should be the identity

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+    map_sentence CASL2Prenex _ _ = error "nyi"
```
tillmo: this should be implemented! see below.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+ in (n', zip newVars $ reverse oldVars)
+
+mapTheory :: (CASLSign, [Named CASLFORMULA]) ->
+          Result (CASLSign, [Named CASLFORMULA])
+mapTheory (sig, nsens) = do
+ let nsens' = map (\nsen -> nsen{sentence =
+      let
+       s = preprocessSen $ sentence nsen
+       (_, s') = computePNF 0 s
+      in s'}) nsens
```
tillmo: this lamda expression looks like `map_sentence`

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+    map_theory CASL2Prenex = mapTheory
+    map_morphism CASL2Prenex = error "nyi"
+    map_sentence CASL2Prenex _ _ = error "nyi"
+    map_symbol CASL2Prenex _ = Set.singleton . id
+    has_model_expansion CASL2Prenex = True -- check
+    is_weakly_amalgamable CASL2Prenex = True --check
+
+-- remove equivalences and implications,
+-- turn arbitrary junctions into binary ones
+preprocessSen :: CASLFORMULA -> CASLFORMULA
```
tillmo: I prefer a prenex normal form computation that leaves the structure of the formula intact as much as possible. Concerning implication and equivalence, see p.10 of http://theo.cs.uni-magdeburg.de/lehre/lehre13w/logik/vl14.pdf
Concerning n-ary conjunctions and disjunctions: they can be handeled analogously to the binary case. So the only needed preprocessing is replacing equivalences by conjunctions of implications.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+              (Negation qsen nullRange) nullRange)
+    _ -> (n', Negation nsen' nullRange)
+  in (n'', nsen'')
+ Junction j [sen1, sen2] _ -> let
+   (n', sen1') = computePNF n sen1
+   (n'', sen2') = computePNF n' sen2
+   (n3, sen') = case (sen1', sen2') of
+     (Quantification q vdecls qsen _, aSen) -> let
+         vars = flatVAR_DECLs vdecls
+         (n''', vars') = getFreshVars vars n''
```
tillmo: I would prefer to make variables disjoint by a global preprocessing. Only in case of a name clash (one and the same variable is used in two different quantifications) a renaming should take place.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+computePNF :: Int -> CASLFORMULA -> (Int, CASLFORMULA)
+computePNF n sen = case sen of
+ Negation nsen _ -> let
+   (n', nsen') = computePNF n nsen
+   (n'', nsen'') = case nsen' of
+    Quantification q vdecls qsen _ -> -- deMorgan
+       (n', Quantification (dualQuant q) vdecls
+              (Negation qsen nullRange) nullRange)
+    _ -> (n', Negation nsen' nullRange)
+  in (n'', nsen'')
```
tillmo: This does not work. Quantifiers may remain inside a negation. E.g.
```
spec Prenex =
  sort s
  preds p,q:s
  . not (exists x:s . ( forall y:s . p(x) \/ q(y)))
end
```
You need to go recursively through `nsen'` and dualise all its quantifiers.

Comorphisms/CASL2Prenex.hs
tillmo: Besides, the current implementation does not work. Distinct variables can be clashed by the translation. E.g.
```
spec Prenex2 =
  sort s
  preds p,q:s
  . (forall x:s . forall y:s . p(x) /\ q(y)) /\ (forall z:s . q(z))
end
```

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+        in (n4, Quantification q vdecls' qsen'' nullRange)
+     (aSen, Quantification q vdecls qsen _) -> let
+         vars = flatVAR_DECLs vdecls
+         (n''', vars') = getFreshVars vars n''
+         vdecls' = map (\(v,s) -> mkVarDecl v s) $ map fst vars'
+         qsen' = foldl (\f (v,s,t)-> substitute v s t f) qsen $
+                        map (\((x,y),(z,t)) -> (z, t, Qual_var x y nullRange))
+                        vars'
+         (n4, qsen'') = computePNF n''' (Junction j [aSen, qsen'] nullRange)
+        in (n4, Quantification q vdecls' qsen'' nullRange)
```
tillmo: these two cases share a lot of code. This can be avoided by treating n-ary junctions in general. Look for the first quantification in the list of con/disjuncts and process it.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,150 @@
...
+    sourceLogic CASL2Prenex = CASL
+    sourceSublogic CASL2Prenex = SL.caslTop --TODO
+    targetLogic CASL2Prenex = CASL
+    mapSublogic CASL2Prenex sl = Just sl -- TODO
+    map_theory CASL2Prenex = mapTheory
+    map_morphism CASL2Prenex = error "nyi"
+    map_sentence CASL2Prenex _ _ = error "nyi"
+    map_symbol CASL2Prenex _ = Set.singleton . id
+    has_model_expansion CASL2Prenex = True -- check
+    is_weakly_amalgamable CASL2Prenex = True --check
```
tillmo: again, since model translation is the identity, both properties hold.

Comorphisms/CASL2NNF.hs
mcodescu: sure, --check means I did not think about it.

Comorphisms/CASL2Prenex.hs
mcodescu: Same as above, I did not think of it.

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,200 @@
...
+    sourceLogic CASL2Skolem = CASL
+    sourceSublogic CASL2Skolem = SL.caslTop --TODO
+    targetLogic CASL2Skolem = CASL
+    mapSublogic CASL2Skolem sl = Just sl -- TODO
+    map_theory CASL2Skolem = mapTheory
+    map_morphism CASL2Skolem = error "nyi"
+    map_sentence CASL2Skolem _ _ = error "nyi"
+    map_symbol CASL2Skolem _ = Set.singleton . id
+    has_model_expansion CASL2Skolem = True -- check
+    is_weakly_amalgamable CASL2Skolem = True --check
```
tillmo: again, since model translation is the identity, both properties hold. But we have a severe problem: the satisfaction condition does not hold. Indeed, the Skolem transformation does not preserve satisfaction in a given model, but only satisfiability. Proposed solution: introduce two more flags, namely eps and rps (expanstion resp. reduction preserves satisfaction, see the terminology of http://matwbn.icm.edu.pl/ksiazki/bcp/bcp28/bcp2815.pdf , there used only for the satisfaction condition of institutiosn). Normally, a comorphism has eps and rps both true, but this one has rps true and eps false, because e.g. M |= alpha(\exists x. P(x)) <=> M |= P(a) => \beta(M)=M|=\exists x. P(x), but not vice versa (\beta plays the role of "reduction" here).

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,200 @@
...
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2Skolem = CASL
+    sourceSublogic CASL2Skolem = SL.caslTop --TODO
+    targetLogic CASL2Skolem = CASL
+    mapSublogic CASL2Skolem sl = Just sl -- TODO
+    map_theory CASL2Skolem = mapTheory
+    map_morphism CASL2Skolem = error "nyi"
+    map_sentence CASL2Skolem _ _ = error "nyi"
```
tillmo: this should be implemented in the same way as for theories

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,200 @@
...
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2Skolem = CASL
+    sourceSublogic CASL2Skolem = SL.caslTop --TODO
+    targetLogic CASL2Skolem = CASL
+    mapSublogic CASL2Skolem sl = Just sl -- TODO
+    map_theory CASL2Skolem = mapTheory
+    map_morphism CASL2Skolem = error "nyi"
```
tillmo: this should extend the given morphisms to the new Skolem functions.

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,200 @@
...
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree
+               CASL CASL_Sublogics
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2Skolem = CASL
```
tillmo: We need to assume that the input is in Prenex form. Counterexample:
```
spec Inconsistent =
  sort s
  pred p:s
  op a:s
  . not (exists x:s . p(x) => p(a))
end
```
This spec is unsatisfiable, but its Skolem translation is satisfiable. Hence, the property that the Skolem translation preserves and reflects satisfiability is lost.
It seems that we need a new sublogic for CASL, namely Prenex form, which should be the source logic here (and the target logic for the Prenex translation). Note that this should be a new constructor between GHorn and FOL (every GHorn formula is Prenex, every Prenex formula is FOL).

tillmo: This is a comment on the source *sub* logic, of course.

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,200 @@
...
+    sourceSublogic CASL2Skolem = SL.caslTop --TODO
+    targetLogic CASL2Skolem = CASL
+    mapSublogic CASL2Skolem sl = Just sl -- TODO
```
tillmo: I think the sublogic stays as it is. `Atom` and `Horn` formulas are already in Skolem form. `Prenex` formulas will stay `Prenex`, and `FOL` is forbidden (see above). The other sublogic features are entierly orthogonal here.

Comorphisms/CASL2Skolem.hs
mcodescu: I think this raises issues: assume we have a theory morphism m from (S, E) to (S', E'). Applying CASL2Skolem to m introduces some Skolem functions in the translation of (S, E) according to E that may have counterparts in the translation of (S', E'). We could work around this by adding the needed symbols to the signature of CASL2Skolem(S', E') and mapping them identically. But I'm thinking this is not how we compute the signature of a node in a dev. graph.

tillmo: yes, indeed there is a problem. If E = { exists x:s . x=x } and E' = \emptyset, then we have a theory morphism. But the Skolemisation, a=a, cannot be mapped, because there is (say) no constant in the target signature.

mcodescu:
I have corrected the code according to your suggestions. What seems to be left is morphism translation for CASL2Skolem and checking that the theory is indeed in CASL.Prenex when applying the same comorphism. I have tested CASL2Prenex on the examples I have from the LPL book, which uses somewhat different rules, and on the ones you have provided in the comments, so if you find (again) mistakes, please let me know.

Comorphisms/CASL2Skolem.hs
```diff
@@ -0,0 +1,207 @@
...
+               CASLMor
+               Symbol RawSymbol ProofTree
+               CASL CASL_Sublogics
+               CASLBasicSpec CASLFORMULA SYMB_ITEMS SYMB_MAP_ITEMS
+               CASLSign
+               CASLMor
+               Symbol RawSymbol ProofTree where
+    sourceLogic CASL2Skolem = CASL
+    sourceSublogic CASL2Skolem = SL.cPrenex
+    targetLogic CASL2Skolem = CASL
```
tillmo: I think the target logic is cPrenex as well, because the transformation will keep Prenex form.

CASL/Sublogic.hs
```diff
@@ -34,6 +34,7 @@ module CASL.Sublogic
     , top
     , caslTop
     , cFol
+    , cPrenex
```
tillmo: OK, but the recognition of the sublogic Prenex does not work yet...

tillmo:
Even if the recognition of the sublogic Prenex does not work yet (see above), I should be able to compoe CASL2Prenex with CASL2Skolem, because the sublogics match. But then selecting "translate theory" at a DG node, this option is not shown.

Comorphisms/CASL2Skolem.hs
mcodescu: But I must give a logic here, not a sublogic of a logic (SL.cPrenex gives an error). The target sublogic is determined by the mapSublogic function (cf its comment in Logic/Comorphism.hs: "finer information of target sublogics corresponding to source sublogics").

mcodescu:
If I change the source sublogic of CASL2Skolem from SL.cPrenex to CASL, the composition CASL2Prenex;CASL2Skolem is available when translating the theory. The problem is indeed the recognition of the sublogic Prenex: the theory resulting after applying CASL2Skolem is not in SL.cPrenex, but in  FOL, and thus CASL2Skolem is not available.

Comorphisms/CASL2Skolem.hs
tillmo: yes, of course

tillmo:
I do not think that th comorphism paths displayed by Hets are computed by translating the theory along all comorphisms in all paths. This would take way too long. Instead, the paths are computed by comparing sublogics, i.e. the target sublogic of the first comorphism must be included in the source sublogic of the second one in order to make them composable. I wonder why this does not work here.

mcodescu:
Right. I have also changed the method mapSublogic of CASL2Prenex: 

mapSublogic CASL2Prenex sl = 
      Just $ sl{SL.which_logic = SL.Prenex}

and thus FOL should be mapped to Prenex (I defined it as cFOL{which_logic = Prenex}). Thus I should get the source logic of CASL2Skolem, and the comorphisms should be composable.

tillmo:
OK, but this may result in a too large sublogic. You should take the intersection of Prenex with the previous value (e.g. Atom then will stay Atom).

tillmo:
I have delegated some issues to #1646 and #1647

tillmo:
With your last version (which is from Tuesday) I cannot test the CASL2Skolem translation, because even after translating with CASL2Prenex, my theories have sublogic FOL.

tillmo:
There is one issue with `CASL2Skolem`: the formula
```
forall x, y : s . forall z : s . (p(x) /\ q(y)) /\ q(z)
```
is translated to
```
forall x : s . (q(gn_sk_f3(x)) /\ p(x)) /\ forall z : s . q(z)
```
That is, the Prenex form is lost. This should not happen.
Moreover,
```
forall x : s . exists y : s . not (p(x) \/ q(y)) 
```
is translated to
```
forall x : s . not (q(gn_sk_f2(x)) \/ p(x))
```
i.e. the disjuncts are swapped. I would prefer to leave the order of disjuncts as it is - this will ease tracing the translation results to the source.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,184 @@
...
+  Quantification q vdecls (computePNF qsen) nullRange
+ Junction j sens _ -> let
+   sens' = map computePNF sens
+   processQuants asen = case asen of
+     Quantification q vdecls sen' _ -> Quantification q vdecls (processQuants sen') nullRange
+     Junction j' jsens _ -> let
+       nonQuantifiedSen (Quantification _ _ _ _) = False
+       nonQuantifiedSen _ = True
+       (nonQuantJSens, remJSens) = span nonQuantifiedSen jsens
+       in if nonQuantJSens == jsens then asen
```
tillmo: I think it is more efficient to test `null remJSens` here

mcodescu: Of course, I have changed it.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,184 @@
...
+        _ -> -- no quantifications
+          Relation sen1' Implication sen2' nullRange
+ Relation sen1 RevImpl sen2 _ -> 
+  computePNF $ Relation sen2 Implication sen1 nullRange
+ Quantification q vdecls qsen _ -> 
+  Quantification q vdecls (computePNF qsen) nullRange
+ Junction j sens _ -> let
+   sens' = map computePNF sens
+   processQuants asen = case asen of
+     Quantification q vdecls sen' _ -> Quantification q vdecls (processQuants sen') nullRange
```
tillmo: this code will never be entered, because `processQuants` is only called on a `Junction`.

Comorphisms/CASL2Prenex.hs
```diff
@@ -0,0 +1,184 @@
...
+     Junction j' jsens _ -> let
+       nonQuantifiedSen (Quantification _ _ _ _) = False
+       nonQuantifiedSen _ = True
+       (nonQuantJSens, remJSens) = span nonQuantifiedSen jsens
+       in if nonQuantJSens == jsens then asen
+           else case remJSens of 
+                  (Quantification q vdecls qsen _):remJSens' -> 
+                    Quantification q vdecls (processQuants $ Junction j' (nonQuantJSens ++ (qsen: remJSens')) nullRange) nullRange
```
tillmo: what I do not like here so much is that the order of sentences in the `Junction` is changed. Moreover, currenlty `span` is called repeatedly, which is rather inefficient. Why not recurse (using `foldl`) through all con/disjuncts and thereby collect all quantifiers, paired with the stripped con/disjuncts, in the accumulator?
Actually, what confuses me a bit is that I cannot observe the re-ordering of conjcuts that I would expect from seeing the code. E.G. consider
```
spec MiexConjunction =
  sort s
  pred p,q:s
  op a,b:s
  . p(a) /\ (forall x:s . p(x)) /\ p(b)  /\ (forall y:s . q(y))
end
```

Comorphisms/CASL2Prenex.hs
mcodescu: Right, I wrote it thinking I'll call the function again for the quantified sentence, then it was not needed to do so and I didn't remove it. Will do so.

Comorphisms/CASL2Prenex.hs
mcodescu: I agree with the refactoring, but I don't see what part of the code should generate the reversal of order in *juncts. The code looks for the first quantification, puts the vdecls in front and leaves the quantified sentence in the same place with the qualification.

mcodescu: I have pushed a new variant that follows your suggestion and gives the expect result for your example above.

Comorphisms/CASL2Prenex.hs
```diff
@@ -155,19 +156,13 @@ computePNF sen = case sen of
...
-                  _ -> error "computePNF" -- should not be possible
-     _ -> asen
-  in processQuants $ Junction j sens' nullRange
+   (vdecls, sens'') = foldl (\(vs, ss) s -> case s of 
+                                           Quantification q vd' qs' _ -> ((q,vd'):vs, qs':ss)
+                                           _ -> (vs, s:ss)) 
+                    ([], []) sens'
+   qsen = Junction j (reverse sens'') nullRange
+   rsen = foldl (\s (q, vd) -> Quantification q vd s nullRange) qsen vdecls 
+  in rsen
```
tillmo: looks much better. However, it seems that looking into one level of quantification is not enough. Consider:
```
spec Mixed =
  sort s
  pred p,q:s
  op a,b:s
  . p(a) /\ ((forall x:s . p(x)) /\ (forall z:s . p(z)))  /\ (forall y:s . q(y))
end
```
One quantifier remains inside.

mcodescu: Pushed a version that collects all quantifiers.

tillmo:
:+1: (the rest is in other issues)

CASL/Sublogic.hs
```diff
@@ -459,7 +459,7 @@ sl_form_level_aux ff (isCompound, leftImp) phi =
        let ql = sl_form_level_aux ff (isCompound, leftImp) f
        in if is_atomic_q q then ql else max FOL ql
    Junction j l _ -> maximum $ case j of
-      Con -> map (sl_form_level_aux ff (True, leftImp)) l
+      Con -> FOL : map (sl_form_level_aux ff (True, leftImp)) l
```
tillmo: This is problematic. The spec
```
spec sp =
  sort s
  preds p,r,q:s
  forall x:s . p(x) /\ q(x) => r(x)
end
```
is now classified as `Prenex`, but it should be `Horn`.

PR #1601, opened 2016-03-15
1587 extracting modules

mcodescu:


tillmo:
Do you have an example that goes through? With
```
logic OWL

ontology O1 = 
  Prefix: : <http://www.omg.org/spec/DOL/DOL-terms/>
  Class: A SubClassOf: B
  Class: B SubClassOf: C
  Class: C
  Class: D
  extract  <http://www.omg.org/spec/DOL/DOL-terms/A> <http://www.omg.org/spec/DOL/DOL-terms/C>
end
```
I get an empty module.

OWL2/ExtractModule.hs
```diff
@@ -0,0 +1,60 @@
...
+
+--import OWL2.AS
+import OWL2.Sign
+import OWL2.MS
+import OWL2.ManchesterPrint
+import OWL2.ParseOWL
+import OWL2.StaticAnalysis
+
+--import qualified Data.Set as Set
+--import qualified Data.Map as Map
```
tillmo: please remove these

OWL2/ExtractModule.hs
```diff
@@ -0,0 +1,60 @@
...
+
+import Common.Utils
+import Common.Result
+import Common.ResultT
+import Common.AS_Annotation
+import Common.GlobalAnnotations
+import qualified Common.IRI as IRI
+
+--import Control.Monad
```
tillmo: please remove this

Static/AnalysisStructured.hs
```diff
@@ -680,7 +680,7 @@ gsigUnionMaybe lg both mn gsig = case mn of
 
 anaExtraction :: LogicGraph -> LibEnv -> DGraph -> NodeSig -> NodeName -> Range ->
               EXTRACTION -> Result (NodeSig, DGraph)
-anaExtraction lg libEnv dg nsig name rg _extr = do
+anaExtraction lg libEnv dg nsig name rg (ExtractOrRemove _ iris _) = do
```
tillmo: if the first argument of `ExtractOrRemove` is `False`, throw an error "remove not implemented yet". This is better than silently ignoring the difference between `extract` and `remove`.

mcodescu:
```
ontology O1 = 
 Prefix: : <http://www.omg.org/spec/DOL/DOL-terms/>
  Class: A SubClassOf: B
  Class: B SubClassOf: C
  Class: C
  Class: D
end

ontology M = 
 O1 extract <http://www.omg.org/spec/DOL/DOL-terms/A>
end
```

goes through.

mcodescu:
Your example above also works as expected for me.

tillmo:
:+1: for now, relegating remaining issues to #1602 and #1603.

PR #1607, opened 2016-03-25
applied utils/replaceAllHeaders.sh

jelmd:
- all $Header$ tokens in doc comments replaced with corresponding filename


tillmo:
:+1:

Issue #1480, opened 2015-03-30
move/replace mathserv


cmaeder:
In SoftFOL.MathServParsing hets calls the MathServerBroker and Vampire using the defaultServer mathserv.informatik.uni-bremen.de on port 8080. It is an old version from 2007 based on java j2sdk1.4.2. Maybe a new service could be set up. 

In any case the service must be moved Bremen to Magdeburg.

cmaeder:
It should be possible to set the environment  variable "HETS_MATHSERVE" to point to a different server and port.

PR #1608, opened 2016-03-29
Makefile enhancements, new pkg names

jelmd:


Makefile
```diff
@@ -439,73 +457,95 @@ gendrifted_files = $(patsubst %.der.hs, %.hs, $(generated_rule_files))
...
+	touch .hets_cgi
+	$(HC) --make -O GUI/hets_cgi.hs -o hets.cgi $(HC_OPTS)
+	ln -f hets.cgi hets_cgi.bin
 
-cgi:
-	$(MAKE) distclean
-	$(MAKE) derivedSources
-	$(MAKE) clean
-	$(MAKE) hets.cgi
+derivedSources: $(derived_sources) $(PFE_SETUP)
```
jelmd: Missed this one: drop $(PFE_SETUP) otherwise distclean would remove it ...

cmaeder:
In order to let travis go through you should remove `parsec1` from `.travis.yml`.

jelmd:
OK, thanx, but how to? Haevn't found any docs, which describe, how to handle such case - only the shallow click-click-click bla.

cmaeder:
:-1: too many warnings #1611

tillmo:
:+1:  (although I have not checked all details)

tillmo:
The sign-off was premature. I have problems with the Makefile in a situation where part of the modules have already been compiled:
```
till@mark:~/projects/Hets$ make
utils/genRules -r ShATermConvertible -i ATerm.Lib -i ATC.Result -o OWL2/ATC_OWL2.der.hs OWL2/AS.hs OWL2/Symbols.hs OWL2/Sign.hs OWL2/MS.hs OWL2/Morphism.hs OWL2/ProfilesAndSublogics.hs OWL2/Sublogic.hs OWL2/Profiles.hs
rm -f OWL2/ATC_OWL2.hs
(DERIVEPATH=; export DERIVEPATH; utils/DrIFT OWL2/ATC_OWL2.der.hs > OWL2/ATC_OWL2.hs)
chmod 444 OWL2/ATC_OWL2.hs
utils/genRules -r ShATermConvertible -i ATerm.Lib -i ATC.Result -o RDF/ATC_RDF.der.hs RDF/AS.hs OWL2/AS.hs RDF/Symbols.hs RDF/Sign.hs RDF/Morphism.hs RDF/Sublogic.hs
rm -f RDF/ATC_RDF.hs
(DERIVEPATH=; export DERIVEPATH; utils/DrIFT RDF/ATC_RDF.der.hs > RDF/ATC_RDF.hs)
chmod 444 RDF/ATC_RDF.hs
touch .hets-oow
ghc -optl-s -XTemplateHaskell -threaded -rtsopts --make -o hets hets.hs   -DTAR_PACKAGE  -DUNIX -DUNI_PACKAGE -DHASKELINE -DHEXPAT -package programatica -DPROGRAMATICA -DSERVER -DWARP3 -DHAXML  -DRDFLOGIC -DCASLEXTENSIONS -DGTKGLADE  -DGTK12 -Wall -fwarn-tabs -fwarn-unrecognised-pragmas -fno-warn-orphans -fno-warn-unused-do-bind 

Haskell/HatAna.hs:19:8:
    Could not find module `Haskell.PreludeString'
make: *** [hets] Fehler 1
```

jelmd:
Can't reproduce.

tillmo:
I use Ubuntu 14.04 with ghc 7.6.3. With a freshly clone repo, I switch to branch `makefile_enhancements`, call `make` and get the above error.

tillmo:
`Haskell/HatAna.hs` is somewhat obsolete, because it uses programatica. The old Makefile avoided compiling it. A better solution would be to move this to a branch.

cmaeder:
try removing `PFE_FLAGS` from `var.mk` first.

tillmo:
yes, this works. But doing this manually is not an option.

cmaeder:
`ghc-pkg unregister programatica` would have the same effect

tillmo:
OK, this works. But I wonder why I had registered this at all. Could some part of Hets have registered this automatically?

cmaeder:
These are the build-deps of our hets package

jelmd:
Ah yes, kind of bug, because the intent is, if programatica/ is empty or points to nirwana, it gets pulled in and prepared automagically. However, it is done for local src (i.e. $PROGRAMATICA_SRC_FILE), only (not for $PROGRAMATICA_SRC_URL). Fix is trivial (i.e. that the latter gets taken into account as well, but still is not 100% safe because fetching it via http may still fail).

But from the comments above I assume, that programatica support is kind of optional, might be dropped?

If so, I would extend the PROGRAMATICAVERSION check, which sets the flag for compiling programatica support, so that `programatica/tools/Setup.hs` is required to exist as well, before this flag get set and adjust the related stuff in the makefile accpordingly.

I would keep the `check-programatica` target (which tries to pull and setup programatica src) as a standalone target for convinience. The other alternative would be to write into README something like "if you want programatica support, pull it from ... and expand to programatica/tools/" and get rid off the related clutter in var.mk/Makefile.

jelmd:
Final changes as suggested above. IMHO that's it.

tillmo:
I found another problem, which does not occur on master
```
till@mark:~/projects/Hets$ touch OWL2/ATC_OWL2.der.hs
till@mark:~/projects/Hets$ make
DERIVEPATH=; export DERIVEPATH; utils/DrIFT OWL2/ATC_OWL2.der.hs > OWL2/ATC_OWL2.hs
/bin/sh: 1: cannot create OWL2/ATC_OWL2.hs: Permission denied
```

cmaeder:
I get:

```
maeder@lati:~/Hets$ touch OWL2/ATC_OWL2.der.hs
maeder@lati:~/Hets$ make
ghc-pkg: cannot find package programatica
rm -f OWL2/ATC_OWL2.hs
(DERIVEPATH=; export DERIVEPATH; utils/DrIFT OWL2/ATC_OWL2.der.hs > OWL2/ATC_OWL2.hs)
chmod 444 OWL2/ATC_OWL2.hs
touch .hets-oow
ghc -optl-s -XTemplateHaskell -threaded -rtsopts --make -o hets hets.hs ...
```

cmaeder:
(so does not fail)

jelmd:
Yes, result of the `chmod 0444 ...`. Wondering, why such a mode gets set? My first idea is: doesn't make sense - drop it completely?

tillmo:
I wonder why I do not get this problem on `master`. The `chmod 0444` is the same there.

jelmd:
But not the `rm -f ...` right before ;-)

cmaeder:
444 was meant to prevent users from changing generated files by hand.

cmaeder:
I had this "rm -f" on makefike_enhancements.

jelmd:
444 - hmmm, very very unusual behavior, i.e. hets make is the first build I've seen, that does this (tries to "protect" generated sources in such a way). The user may always get rid off it easily, however, seemed to work for Till's case :). Anyway, still opt for dropping it - means also less clutter when building it ...

jelmd:
After the last change `rm -rf ...` should only occur within the clean* and Version.hs targets ...

cmaeder:
Well, remove `chmod 444 $@` if it hurts. The generated files contain two `don't touch` warnings.

jelmd:
OK, thanx.

tillmo:
:+1:

jelmd:
Is there a way to submit tests directly to whatever machine is doing the tests above?
(Don't wanna pollute this branch with try-and-error stuff ...).

cmaeder:
simply create a new branch and test there

jelmd:
Dann bitt ich noch mal um 'nen Daumen. travis CI rennt jetzt auf trusty statt precise, braucht zwecks native packages zwischen 15..20 min weniger Zeit und führt dabei auch noch die spass und darwin relevante tests aus (scheinen vorher immer still fehlgeschlagen zu sein).

tillmo:
when looking at your changes, I noticed that `texlive-common` etc. is a dependency. I think this should only be recommended. One can very well use Hets without having LaTeX installed. It is just that one of Hets' output options produces tex.

jelmd:
Hmmm, can't find it. The only place where texlive-* occurs is in the 'Build-Depends:` section of `debian/control`. It is required to build the UserGuide.pdf  (and for tests as well) ...

tillmo:
yes, I meant `debian/control`. I would make `texlive` recommended there. For buildung the user guide and the tests, it needs to be installed.

jelmd:
Yes, that's the purpose of 'Build-Depends:' - all stuff listed here gets automatically installed if debian tools are used to build it or if one does a `apt-get build-dep hets`. It applies to the source package, only.

We could add a [Recommends | Suggests](https://www.debian.org/doc/debian-policy/ch-relationships.html#s-binarydeps) section to the binary package definitions with all the 'texlive-*` packages, but than the question is, where to start and where to stop! IMHO such sections don't make much sense, because AFAIK no other pkg system supports such "soft" dependencies and therefore even if we can make the same packages wrt. name and contents, the install instructions would need to be re-written for each pkg system =8-(.

Adding them to the _hets-{desktop|server}-all_ `Depends:` however would make sense IMHO.

BTW: Is graphviz (dot) a hard dependency as well?

tillmo:
oh, a `texlive` dependency for the source package is fine. Where do I find the dependencies for the binary packages?

jelmd:
Each `^Package:` starts a new binary package description.

jelmd:
BTW:
Is anyone using `./hets.in`. It seems to be an artifact as well as `./doc/BuildingAnInstaller.tex`.
Should be removed?

`./clean.{lst,sh}`  are also nowhere referenced. Should I weave them into the Makefile as target `really_really_clean` and remove too?

cmaeder:
> BTW: Is graphviz (dot) a hard dependency as well?

graphviz is needed to produce svg graphs within the web interface

cmaeder:
> Is anyone using ./hets.in.

I think that was used with our old izpack installers that we no longer use and can go away.

`./clean.{lst,sh}` was used to create a (cleaned-up) source distribution which we also no longer provide.

jelmd:
graphviz: Yepp, I know. But does it fall into the "can very well use Hets without having ... installed" category?
Or  will hets misbehave/coredump/etc if it is not installed? Is it essential in the day-by-day workflows? If not, it should be handled like the TeX stuff ...

hets.in,..., clean*: I count it as a 1+ for removal.  BTW: One may consider the archive made by "make archive" as a cleaned up source distribution - I'm pushing this as orig source to PPA (previously 2x ~69 MB, now 1x ~21 MB). However, thought that the docs and test might be useful as well, so kept it as is. Removing them would save ~3 MB wrt. archive, ~9 MB if extracted (~800 files).

tillmo:
:+1:

jelmd:
Hmmm, seems like the Vampire test hangs and screws up testing =8-(((
I guess, proper connection timeout handling needs to be implemented ...

cmaeder:
> graphviz: Yepp, I know. But does it fall into the "can very well use Hets without having ... installed" category?
> Or will hets misbehave/coredump/etc if it is not installed? Is it essential in the day-by-day workflows?

I don't know. The code is 

```
getSVG :: String -> String -> DGraph -> ResultT IO String
getSVG title url dg = do
        (exCode, out, err) <- lift $ executeProcess "dot" ["-Tsvg"]
          $ dotGraph title False url dg
        case exCode of
          ExitSuccess -> liftR $ extractSVG dg out
          _ -> fail err
```

it is not used, if you only want xml from hets

cmaeder:
> seems like the Vampire test hangs and screws up testing

Vampire is only called via the MathServe Broker. Was this ever moved to Magdeburg?

jelmd:
Vampire: Nope - missing source code prevents porting/adaption to a recent OS/JDK.

jelmd:
dot: Thanx. Till said, we keep it as is for now - should perhaps replace udraw* some day ...

cmaeder:
maybe mathserve should be retired (see #1480)

jelmd:
Yupp. At least using `export HETS_MATHSERVE=mathserv.iws.cs.ovgu.de:8080` would not block.
Anyway, no matter which server is used, hets should be immune against server malfunctioning.
_callMathServ_ says something about a time limit. Not sure, whether this can be used ...

BTW: travis people said, that jobs can be restarted, if one logs in ... Does anyone have these data, can restart the last failed checks?

cmaeder:
Yes, I've just restarted it. https://travis-ci.org/spechub/Hets (after "sign in with github")

Issue #1412, opened 2014-11-18
fact++ package under ubuntu


cmaeder:
Hets uses fact++ version 1.6.1. The dynamic libraries needed by hets are supplied by the package `libfactplusplusjni` that is part of our own fact++ package. On utopic our fact++ package could be installed using the exact version by `sudo apt-get install fact++=1.6.1-0ubuntu7`

Ubuntu's newer version (1.6.2) does not supply `libfactplusplusjni`! The actual binary `FaCT++` is not used by hets. So the right dependency is best installed by `sudo apt-get install libfactplusplusjni`. This dependency should be re-added to hets-ontology.

The actual problem is, that hets reports Fact as available prover without checking the dynamic library (`/lib/libFaCTPlusPlusJNI.so`) supplied by `libfactplusplusjni`.

tillmo:
> Maybe the dependency `libfactplusplusjni` should be added to hets-core (and hets-server-core).

Yes please!

jelmd:
Perhaps as recommended because it does not really depend on it?

cmaeder:
It really depends, as it can be called and will crash. The alternative is to make hets smarter to recognized that it cannot run fact.

tillmo:
So for now, it should depend on it. Later on, we may switch this to recommended.

jelmd:
Ah ok. But wait, AFAICS hets already ships its own version below hets/hets-owl-tools/lib/{uk.ac.manchester.cs.owl.factplusplus-P5.0-v1.6.3.1.jar,/native/*/}

cmaeder:
Yes, `libfactplusplusjni` additionally supplies native dynamic libraries that are needed, too.

cmaeder:
Well, maybe the proper file from https://github.com/spechub/Hets/tree/master/OWL2/java/lib/native
can be copied to the correct location by the hets installation itself. Currently this is separated.

jelmd:
Ehmm, basically the JNI lib of fact++ is the same as the binary FaCT++, but without the main method and with the dyn lib/loadable JNI compat, interface, so that java app is able to call it. So if hets doesn't use it by itself, I think, just the proper environment variable needs to be set or param given to the jvm invokation, to pick it up automagically.

cmaeder:
Currently, hets calls OWLFact.jar and OWLFactProver.jar that are created from the java sources under https://github.com/spechub/Hets/tree/master/OWL2/java/de/unibremen/informatik

[...]

PR #1619, opened 2016-04-20
Add support for Fact++ jnilibs files.

eugenk:
This should allow to use both `libFaCTPlusPlusJNI.jnilib` and `libFaCTPlusPlusJNI.so` with Hets for Fact++. For the `libFaCTPlusPlusJNI.jnilib`, the `HETS_JNI_LIBS` environment variable is supposed to point to the directory containing the `libFaCTPlusPlusJNI.jnilib`.

This is about the discussion in #1412.

### DIFF
--- a/OWL2/ProveFact.hs
+++ b/OWL2/ProveFact.hs
@@ -161,8 +161,6 @@ runTimedFact tmpFileName prob mEnt tLimit = do
       jlibSoName = "libFaCTPlusPlusJNI.so"
       jlibName = "libFaCTPlusPlusJNI.jnilib"
   (progTh, toolPath) <- check4HetsOWLjar jar
-  hasJniSoInLib <- doesFileExist $ "/lib/" ++ jlibSoName
-  hasJniInLib <- doesFileExist $ "/lib/" ++ jlibName
   (_, arch, _) <- executeProcess "uname" ["-m"] ""
   if progTh then
         withinDirectory toolPath $ do
@@ -170,13 +168,12 @@ runTimedFact tmpFileName prob mEnt tLimit = do
           let jni = fromMaybe ("lib/native/" ++ trim arch) mJni
           hasJniSo <- doesFileExist $ jni </> jlibSoName
           hasJni <- doesFileExist $ jni </> jlibName
-          if hasJni || hasJniSo || hasJniInLib || hasJniSoInLib then do
+          if hasJni || hasJniSo then do
             timeTmpFile <- getTempFile prob tmpFileName
             let entailsFile = timeTmpFile ++ ".entail.owl"
-                jargs = ["-Djava.library.path=" ++ jni
-                        | not (hasJniInLib || hasJniSoInLib) || isJust mJni ]
+                jargs = ["-Djava.library.path=" ++ jni | isJust mJni ]
                    ++ ["-jar", jar, "file://" ++ timeTmpFile]
-                  ++ ["file://" ++ entailsFile | hasEnt ]
+                   ++ ["file://" ++ entailsFile | hasEnt ]
             case mEnt of
               Just entail -> writeFile entailsFile entail
               _ -> return ()
--- a/OWL2/ProveFact.hs
+++ b/OWL2/ProveFact.hs
@@ -11,6 +11,7 @@ Fact++ prover for OWL
 -}
 
 module OWL2.ProveFact (factProver, factConsChecker) where
+import Data.List (intercalate) -- DEBUG
 
 import Logic.Prover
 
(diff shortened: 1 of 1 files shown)


OWL2/ProveFact.hs
```diff
@@ -158,34 +159,46 @@ runTimedFact :: FilePath -- ^ basename of problem file
...
+          putStrLn $ "mLibraryPath: " ++ show mLibraryPath -- DEBUG
+          executeProcess "cp" [timeTmpFile, "/tmp/timeTmpFile.owl"] "" -- DEBUG
+          executeProcess "cp" [entailsFile, "/tmp/entailsFile.owl"] "" -- DEBUG
+          t_start <- getHetsTime
+          mExit <- timeoutCommand tLimit "java" jargs
+          putStrLn $ (\ (Just (ex, out, err)) -> "exitCode: " ++ show ex) mExit -- DEBUG
+          putStrLn $ (\ (Just (ex, out, err)) -> "out: " ++ out) mExit -- DEBUG
+          putStrLn $ (\ (Just (ex, out, err)) -> "err: " ++ err) mExit -- DEBUG
+          putStrLn $ "" -- DEBUG
+          putStrLn $ "" -- DEBUG
```
tillmo: probably the DEBUG stuff should go

tillmo:
Actually, when trying to prove with Fact, I get
``` 
Java Error:
FaCT++.Kernel: Reasoner for the SROIQ(D) Description Logic, 64-bit
Copyright (C) Dmitry Tsarkov, 2002-2011. Version 1.5.2 (12 April 2011)
Exception in thread "main" java.lang.UnsatisfiedLinkError: uk.ac.manchester.cs.factplusplus.FaCTPlusPlus.setFreshEntityPolicy(Z)V
	at uk.ac.manchester.cs.factplusplus.FaCTPlusPlus.setFreshEntityPolicy(Native Method)
	at uk.ac.manchester.cs.factplusplus.owlapiv3.FaCTPlusPlusReasoner.<init>(Unknown Source)
	at de.unibremen.informatik.FactProver.FactProver.<init>(FactProver.java:98)
	at de.unibremen.informatik.FactProver.Prove.main(Prove.java:15)
```
This happens also without this pull request. However, in order to test it, I need Fact to work.

My configuration is:
```
apt-cache policy libfactplusplusjni
libfactplusplusjni:
  Installiert:           1.6.3-0ubuntu2
  Installationskandidat: 1.6.3-0ubuntu2
  Versionstabelle:
 *** 1.6.3-0ubuntu2 0
        500 http://ppa.launchpad.net/hets/hets/ubuntu/ trusty/main amd64 Packages
        100 /var/lib/dpkg/status
```

jelmd:
Actually /usr/lib/hets/hets-owl-tools/lib/native/x86_64/ should contain libFaCTPlusPlusJNI.so and thus the libfactplusplusjni package is redundant, i.e. not used by hets. However, it's C++ stuff ...
So checking, whether it gets picked up and if so, renaming it to something else (or symlinking it to the /lib/libFaCTPlusPlusJNI.so) and try again may help to get a clue, what's going on ...

jelmd:
BTW: Just remembered:In early days (depending on the available RAM) java invoked per default the 32bit JVM, i.e.  if not explicitly specified via -d{32|64}. So this may clash as well: java starts the 32bit JVM but hets "misleaded" by `uname -a` == x86_64 say, it should load the 64bit lib ...

So IMHO best is to start the JVM always with -d64 and just load the 64bit lib. Having this in the back of the head as well as the fact, that hets often uses a lot of memory > 2GB, it makes sense to drop 32bit support completely ...

jelmd:
Another thing: Wondering whether [the rlative path](https://github.com/spechub/Hets/blob/factpp_jnilibs_support/OWL2/ProveFact.hs#L167) gets resolved relative to ${HETS_OWL_TOOLS}?

Also just checking LD_LIBRARY_PATH is wrong, because it ignores LD_LIBRARY_PATH_{32|64}. I.e. if LD_LIBRARY_PATH_32 is set, the OS link-editor ignores LD_LIBRARY_PATH completely for 32bit binaries, if LD_LIBRARY_PATH_64 is set,  it ignores LD_LIBRARY_PATH completely for 64bit binaries. So basically LD_LIBRARY_PATH is more or less a bad fallback, since it may screw up things as well because of mixing up 32 and 64bit binaries ...

So if one uses a distro, which provides 32bit as well as 64bit binaries side by side (there is nothing wrong  with it), if needed one should set e.g. LD_LIBRARY_PATH_32=/local/usr/lib ; LD_LIBRARY_PATH_64=/local/usr/lib/x86_64 and unset LD_LIBRARY_PATH to avoid getting hurt by such flawed scripts, ...

tillmo:
I do not have 32 bit Java, so that is not the cause of the problem:
```
java -d32
Error: This Java instance does not support a 32-bit JVM.
Please install the desired version.
```
I would like to have solution for my problem...

eugenk:
I reverted the debug commit. I'm not sure if this is done. Since this is quite old, I don't know how to test if Fact++ is used properly. @tillmo, do you have an example for me?

tillmo:
for example, https://ontohub.org/esslli-2016/OWL/Family_structured.dol

eugenk:
This is not needed any more. It works just fine on master with macOS.

tillmo:
indeed, things do not work _with_ the pull request!

Issue #1609, opened 2016-04-01
Not all axioms are exported via Hets server


mcodescu:
The graph obtained when calling ```curl localhost:8000/dg/http://148.251.85.37:8300/static/hetsfile10935.dol ``` has just 2 axioms for the node Generalisation0, but in the specification there are 5 axioms. This is also visible here http://ontohub.org/sandbox/hetsfile10935//Generalisation0///axioms

tillmo:
@cmaeder, do you have an idea here?

cmaeder:
I  cannot analyse this file

```
maeder@lati:~/Hets$ ./hets -g hetsfile10935.dol 
Analyzing file hetsfile10935.dol as library hetsfile10935
Analyzing spec hetsfile10935?Atom
Analyzing spec hetsfile10935?Solar
Analyzing spec hetsfile10935?Generalisation0
/home/maeder/Hets/hetsfile10935.dol:80.54: *** Error:
unexpected mixfix token: <
/home/maeder/Hets/hetsfile10935.dol:81.18: *** Error:
unexpected mixfix token: >
/home/maeder/Hets/hetsfile10935.dol:82.39: *** Error:
unexpected mixfix token: >
hets: user error (Stopped due to errors)
```

which branch am I supposed to use?

cmaeder:
After using `<` and `>` as prefix operators, I see 5 axioms in Generalisation0 (also in my xml output)

```
. <=(0, 0) %(Ax1)%

forall G_G910393 : Int . <=(0, G_G910393) => <=(0, suc(G_G910393))
                                                           %(Ax2)%

forall G_G908948, G_G908857 : Int
. <(G_G908857, G_G908948) <=> >(G_G908948, G_G908857)      %(Ax3)%

. >(mass(G_G907993), mass(G_G907842)) %(Ax4)%

. >(force(G_G913050, G_G907842, G_G907993), 0) %(Ax5)%
```

cmaeder:
```
curl localhost:8000/dg/file%3A%2f%2f%2fhome%2fmaeder%2fHets%2fhetsfile10935.dol
```

with a running server produces the same output as `./hets -v2 -o xml hetsfile10935.dol`

cmaeder:
I could reproduce your error for your original wrong `hetsfile10935.dol`! The hets server seems to ignore the parsing error silently, but don't ask me why.

cmaeder:
Maybe errors are not "forced" (aka evaluated).

[...]

Issue #1596, opened 2016-03-14
use IRIs uniformly in HETS


mcodescu:
We currently have IRIs defined in OWL2, in Common.IRI and moreover we have CASL identifiers. This complicates things, e.g. when translating from OWL to CASL or when using Common.IRIs for OWL2. Ideally, there should be just one type of identifiers, that we could create by extending Common.IRI with some mixfix annotations to cover CASL identifiers as well. We should also have a convention, documented in the wiki, about implicit values for fields in the IRI type. As a result, all logics (including CASL and OWL2) should use the new Common.IRI type.

Hets modules to look at: `Common.IRI` (datatype `IRI`), `OWL2.AS`  (datatype `QName`, same as `IRI`, but different from `IRI` in  `Common.IRI`), `Common.Id` (datatype `Id`). These need to be integrated into a uniform datatype. Later on, look at the parsers:  `Common.IRI`, `OWL2.Parse`, `Common.Token` (e.g. `parseId`).

tillmo:
For mapping CASL IDs to IRIs, if there is no empty prefix, we need a default. This could be the IRI of the library and should be consistent with Ontohub. What happens with fully qualified names `f:s->t`, infixes `__+__` and compound IDs `List[Elem]`? First idea: just include them verbatim. Does the IRI syntax allow this?

tillmo:
The problem is that in the standard http://www.ietf.org/rfc/rfc3987.txt brackets `[]` are forbidden in IRIs, while some symbols like `+` and `:` are allowed (at least in the local part). However, all unicode characters beyond ASCII are allowed, too. Hence we could replace square brackets by some unicode brackets. Alternatively, round brackets `()` are allowed, too (but these are already used for application to arguments), as well as curly braces `{}` (but these already are used for DOL structured OMS).

tillmo:
For mapping CASL IDs to IRIs, if there is no empty prefix, we need a default. This could be the IRI of the library and should be consistent with Ontohub. However, in Ontohub, we need to disambiguate between OMS, mappings, symbols, axiom names etc. We could append a keyword `OMS`, `symbol` etc. to the IRI, which is quite unusual, but from a Hets perspective this would work better than prepending it, because Hets needs to work not only with Ontohub, but also with other sources on the web.

tillmo:
[...]

Issue #1565, opened 2016-02-08
forbid axiom labels that are equal to symbols


tillmo:
for the need of this see https://github.com/ontohub/ontohub/issues/1585
Altermatively, rename labels (e.g. by adding "_1") until they are distinct to symbols. This should also be done for unions that lead to clashes.

mcodescu:
I will forbid labels that overlap with symbol names.

tillmo:
I think we do not need this anymore in Ontohub 2.0.

PR #1630, opened 2016-05-03
fail if names of sentences appear as names of symbols

mcodescu:
Without the call of `nameAndDisambiguate`, the following specification goes through:

```
spec S = 
 sort s_1
 op c : s_1
 op f : s_1 -> s_1
 
 . f(c) = c %(s)%
 . exists x : s_1 . f(x) = x %(s)%
end
```

but the second sentence is renamed to `s_1` for proving.

### DIFF
--- a/Logic/Logic.hs
+++ b/Logic/Logic.hs
@@ -394,6 +394,16 @@ symset_of :: forall lid sentence sign morphism symbol .
              lid -> sign -> Set.Set symbol
 symset_of lid sig = Set.unions $ sym_of lid sig
 
+-- | check that all sentence names in a theory do not appear as symbol names
+checkSenNames :: forall lid sentence sign morphism symbol . 
+                 Sentences lid sentence sign morphism symbol =>
+                 lid -> sign -> [Named sentence] -> Set.Set String
+checkSenNames lid aSig nsens = 
+ let senNames = map senAttr nsens
+     symNames = map (show . (sym_name lid)) $ Set.toList $
+                symset_of lid aSig
+ in Set.intersection (Set.fromList symNames) $ Set.fromList senNames 
+
 -- | dependency ordered list of symbols for a signature
 symlist_of :: forall lid sentence sign morphism symbol .
               Sentences lid sentence sign morphism symbol =>
--- a/PGIP/Server.hs
+++ b/PGIP/Server.hs
@@ -314,8 +314,8 @@ oldWebApi opts tempLib sessRef re pathBits splitQuery meth respond
                  Just k' -> getHetsResult opts [] sessRef (qr k')
                               Nothing OldWebAPI proofFormatterOptions
                respond $ case ms of
-                 Nothing -> mkResponse textC status422 $ showRelDiags 1 ds
-                 Just (t, s) -> mkOkResponse t s
+                 Just (t, s) | not $ hasErrors ds-> mkOkResponse t s
+                 _ -> mkResponse textC status422 $ showRelDiags 1 ds
            -- AUTOMATIC PROOFS E.N.D.
            else getHetsResponse opts [] sessRef pathBits splitQuery respond
       "POST" -> do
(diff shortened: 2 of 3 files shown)


mcodescu:
Should solve #1565

tillmo:
OK, but I found some subtle problem. Namely, the following clash is not detected:
```
spec sp =
  sort s
  . forall x:s. x=x %(s1)%
then
  sort s1
end
```

mcodescu:
I was thinking it's too much to call checkSpecNames for a specification involving arbitrary structuring, because this would require to compute its theory and this may be expensive. But I don't know what rules apply regarding the loc/ids of the sort and the sentence s1.

tillmo:
I will discuss this at the Ontohub meeting tomorrow.

mcodescu:
```
spec sp =
  sort s
 then
  . forall x:s. x=x %(s)%
```

This is currently treated as a naming error of the sentence, but if your example above should be correct, then this example should also be. I'm thinking maybe we need internal name of symbols in Hets such that the two symbols `s` in the specification above are distinguished, and these names should be compared. Note that this also requires a generic notion of symbol names in Hets, common to all logic or generalizing specific solutions like CASL Id, OWL IRIs and Common.IRI. This is related to #1596.

eugenk:
This works fine with the command line interface:
```
 ❯ hets -o xml --full-signatures -a none -v2 -O /tmp/ +RTS -K1G -RTS --full-theories -A spec/fixtures/ontologies/clif/cat.clif
Reading file spec/fixtures/ontologies/clif/cat.clif
Downloaded spec/fixtures/ontologies/clif/cat.clif
Analyzing library <Cat>
logic CommonLogic
Analyzing spec <Cat>
*** Error:
The following sentence names are also used as symbol names:Cat
hets-bin: user error (Stopped due to errors)
```

But doesn't work with the REST API:
```
 ❯ curl -s -D - http://localhost:8000/dg/file%3a%2f%2f%2fUsers%2feugen%2fWork%2fOntohub%2fontohub%2fspec%2ffixtures%2fontologies%2fclif%2fcat%2eclif/full-signatures/full-theories/auto\?input-type\=clif\;
HTTP/1.1 200 OK
transfer-encoding: chunked
Date: Mon, 09 May 2016 07:44:04 GMT
Server: Warp/3.1.12
Content-Type: application/xml

<?xml version='1.0' ?>
<DGraph filename="/Users/eugen/Work/Ontohub/ontohub/spec/fixtures/ontologies/clif/cat.clif" mime-type="text/clif" libname="Cat" dgnodes="0" dgedges="0" nextlinkid="0" />
```
At lease we want the same error message and an error status code (instead of 200).

mcodescu:
This seems to be the same issue as what Christian mentioned for #1609, I'll try to fix it.

mcodescu:
I have pushed a change that gives me the following error for wrong names:
```
curl  http://localhost:8000/dg/file%3a%2f%2f%2fhome%2fmcodescu%2frepos%2fHets2%2fstruct%2edol/full-signatures/full-theories/auto
/home/mcodescu/repos/Hets2/struct.dol:2.3-5.1: *** Error:
The following sentence names are also used as symbol names:s
```

@eugenk is this what you expect?

eugenk:
That's exactly what I expect. Thanks!

eugenk:
From my point of view as a user, this is fine. I can't review the code, however, because I don't know the surrounding code.

tillmo:
we will leave the solution of the structuring problems to Ontohub. Hence, this solution is fine.
:+1:

mcodescu:
@eugenk could you please document why this commit should be reverted? I don't have any problem with the revert itself, but we should keep track of these changes.

eugenk:
Oh, yes, of course!

The changes of this pull request are fine for "Legacy Ontohub", where the namespace for symbols is the same as the namespace for sentences (and other parts). However, in the re-implementation of Ontohub, we will have different namespaces for symbols, sentences, mappings etc.

Secondly, in many files of the TPTP library, a sentence is named like a symbol in that sentence (e.g. if a sentence is defining a symbol `s`, this sentence is named `s` as well. Legacy Ontohub catches these errors anyway because of https://github.com/spechub/Hets/pull/1630#issuecomment-217741712 and we should remove it for the sake of Ontohub's re-implementation.

PR #1642, opened 2016-06-01
1620 owl http redirects

### LINKED ISSUE #1620: avoid parsing HTML redirections
The call
```
hets http://ontohub.org/ref/17/meta/dol-ontology
```
leads to a parsing error, because this URL yields a 302 redirect (to https://ontohub.org/ref/17/meta/dol-ontology) and Hets tries to parse the HTML of the redirection. Instead, Hets should follow the redirection(s) and only then parse the output.

jelmd:
fixes #1620

### DIFF
--- a/OWL2/java/de/unibremen/informatik/OWL2Parser.java
+++ b/OWL2/java/de/unibremen/informatik/OWL2Parser.java
@@ -1,4 +1,3 @@
-
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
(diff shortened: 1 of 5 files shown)


OWL2/java/de/unibremen/informatik/OWL2Parser.java
```diff
@@ -45,20 +48,87 @@
...
+	private static URLConnection getConnection(URL url, int redirCount)
+		throws IOException
+	{
+		URLConnection c = url.openConnection();
+		if (!(c instanceof HttpURLConnection)) {
+			return c;
+		}
+		HttpURLConnection con = (HttpURLConnection) c;
+		con.addRequestProperty("Accept", "text/plain");
+		con.setConnectTimeout(300 * 1000);	// 5 min timeout
```
tillmo: couldn't this timeout lead to problems for very large ontologies that take longer than 5 min?

OWL2/java/de/unibremen/informatik/OWL2Parser.java
```diff
@@ -45,20 +48,87 @@
...
+		return con;
+	}
 
 	public static void main(String[] args) {
 		// A simple example of how to load and save an ontology
 		try {
 			OWLOutputHandler out = new OWLOutputHandler();
 			parseArgs(args, out);
-			URLConnection con = uri.toURL().openConnection();
-			con.addRequestProperty("Accept", "text/plain");
```
tillmo: why have you removed the configuration of the accept header?

jelmd: It got moved to line 78 in getConnection() because we don't know in advance, wether we get a redirect and thus we need to pass it for intermediate connects as well.

OWL2/java/de/unibremen/informatik/OWL2Parser.java
jelmd: No.

This is not related to "download time". It is related to the tcp connection (initial handshake) setup and basically says: if a connection can't be established within 5 min, stop waiting for an answer and fail. Therefore I think 5 min are more than sufficient.

Another tweak would be the readTimeout setting, i.e. how long max. a single read on the related socket may block. This would go a little bit more in the "server too busy to answer" direction (IIRC the default is "wait forever"). So setting both to lower values is more or less a resource optimization, higher values usually make not much sense.

Anyway, therefore I think, "download size/time" is kind of orthogonal to these settings.

tillmo:
:+1:

PR #1643, opened 2016-06-02
1641 prefixes of modules

mcodescu:
Use expandedIRI when translating to CASL and unite the prefix map of the module with the one of the ontology it is extracted from, preferring the latter.


tillmo:
:+1:

PR #1661, opened 2016-08-08
relax upper bounds

cmaeder:
I want to merge in this Hets.cabal change so that `cabal install --only-dependencies` works (after glade is installed).


cmaeder:
this change is rather uncritical. I'm not sure, if I should simply merge it or if I should wait for a :+1:

tillmo:
:+1:

PR #1664, opened 2016-08-12
Use ggrep if it exists.

### LINKED ISSUE #1663: grep -e does not work on Solaris
https://github.com/spechub/Hets/blob/master/SoftFOL/ProveDarwin.hs uses `grep -e`. This leads to problems under Solaris, see https://github.com/ontohub/ontohub/pull/1772

solution: use ggrep on Solaris.

eugenk:
This shall fix #1663.


tillmo:
:+1:

PR #1667, opened 2016-08-13
reintroduce fix from pull request #1486

tillmo:
new try of solving #1485 (after @jelmd has re-introduced this bug...)


eugenk:
:+1:

PR #1668, opened 2016-08-14
Fixing owl22casl

mcodescu:
Before the datatype translation is fixed, we should merge the concept and sentence translation. A test file is here: https://ontohub.org/sandbox/owl22casl.dol but if I missed something let me know. Also, if you don't like to have all these generated names for variables, I could implement another convention.

### DIFF
--- a/CASL/Induction.hs
+++ b/CASL/Induction.hs
@@ -12,7 +12,7 @@ We provide both second-order induction schemes as well as their
 instantiation to specific first-order formulas.
 -}
 
-module CASL.Induction (inductionScheme, generateInductionLemmas) where
+module CASL.Induction (inductionScheme, generateInductionLemmas, substitute) where
 
 import CASL.AS_Basic_CASL
 import CASL.Sign
--- a/CASL_DL/PredefinedCASLAxioms.hs
+++ b/CASL_DL/PredefinedCASLAxioms.hs
@@ -286,10 +286,8 @@ predefinedAxioms = let
             (Qual_pred_name thing classPredType n) [t1] n]
 
 mkNNameAux :: Int -> String
-mkNNameAux k = case k of
-    0 -> ""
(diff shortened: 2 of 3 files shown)


tillmo:
The translation of `max` is not correct. For input
```
logic OWL
spec sp =
  Class: A
  ObjectProperty: r  
  Class: B SubClassOf: r max 2 A
with logic OWL22CASL
end
```
I get
```
forall gn_x1 : Thing
. B(gn_x1)
  => forall gn_x2 : Thing; gn_x3 : Thing; gn_x4 : Thing
     . r(gn_x1, gn_x2) /\ r(gn_x1, gn_x3) /\ r(gn_x1, gn_x4) /\ A(gn_x2)
       /\ A(gn_x3) /\ A(gn_x4)
       => gn_x4 = gn_x2 \/ gn_x4 = gn_x3  
```
But this implies that there is at most 1 such element. Take gn_x2 = gn_x3, then every gn_x4 has to be equal to one of them (and therefore both).

tillmo:
Note that the translation of the same ontology with `exactly` instead of `max` is correct, because there, `not gn_x2 = gn_x3` is assumed.

mcodescu:
This should be: 

```
forall gn_x1 : Thing
. B(gn_x1)
  => forall gn_x2 : Thing; gn_x3 : Thing; gn_x4 : Thing
     . r(gn_x1, gn_x2) /\ r(gn_x1, gn_x3) /\ r(gn_x1, gn_x4) /\ A(gn_x2)
       /\ A(gn_x3) /\ A(gn_x4)
       => gn_x4 = gn_x2 \/ gn_x4 = gn_x3 \/ gn_x2 = gn_x3 
```
namely, if the property holds for more than 2 elements, then at least two of them are equal. Am I right?

tillmo:
yes, right

mcodescu:
done, see last commit

tillmo:
:+1: modulo travis

PR #1674, opened 2016-08-16
DOL mode for emacs, based on CASL mode

tillmo:


tillmo:
@mcodescu, and this one: basically please just try whether it works.

mcodescu:
I have opened about 20 dol files with the emacs mode on, and the syntax highlighting works as expected (then, with, combine, alignment, %implies and so on). If some issue shows up, it can be easily corrected later. :+1:

PR #1681, opened 2016-08-19
Fix the version timestamp

eugenk:
If no file is changed in a commit, `git log -1 -- .` does not display
this commit. The log of the whole repository must be used. Therefore,
the `-- .` must be removed.

Before, `make rev.txt` did not consider the merge commit b259e3b.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ define EOL
 endef
 
 # indicate, whether working on an exported repo
-GIT_TIMESTAMP := $(shell [ -d .git ] && git log -1 --format=%ct -- . )
+GIT_TIMESTAMP := $(shell [ -d .git ] && git log -1 --format=%ct )
 EXPORTED := $(shell [ -n "$(GIT_TIMESTAMP)" ] || printf 1)
 
 # the 'replacing spaces' example was taken from the (GNU) Make info manual


jelmd:
Doesn't make any sense to me. BTW: if no files have changed, git responds with something like "... nothing to commit" on commit ...

eugenk:
> if no files have changed, git responds with something like "... nothing to commit" on commit ...

True, but merge commits don't change any files (given there are no conflicts).

> Doesn't make any sense to me

Please try it locally: ;)

* check out b259e3b3e05433b2018f45d5781000fa3af7cbdc (current master)
* compare the output of `git log -1` and `git log -1 -- .` and see which command points to the checked out commit.

jelmd:
IC, you are referring to -no-ff merge commits, which usually just contain meta-data about the merge and you wanna have them taken into account, even so they contain no changes to the codes base.

Hmmm, the initial idea was, that basically only codes base changes matter and a possibly set GIT_DIR gets "overwritten".  However, I don't prefer any, because - as already said - it is a big flaw to use any VCS meta data for versioning/version validation.

tillmo:
Considering merge commits is necessary, because it may be the case that they merge older code into (and thus change) the code base. This would not be discovered when just using the latest code-changing commit. So :+1:

Issue #1556, opened 2016-01-08
Ubuntu packages hets and hets-server


eugenk:
How and when are the Ubuntu packages `hets` and `hets-server` created and rolled out?
Is this a fully automated process?
Does this happen every time the master branch updates?

cmaeder:
I do this manually, ideally only by calling https://github.com/spechub/Hets/blob/master/utils/debian/auto-package/package_trunk (or package_server) with suitable parameters (i.e. `-s wily create`).

I noticed, however, that currently the file `rev.txt` is badly overwritten by the ppa build:
https://launchpadlibrarian.net/233344982/buildlog_ubuntu-wily-amd64.hets-server_0.99r1452189354-0ubuntu1~wily_BUILDING.txt.gz

```
make[2]: git: Command not found
rm -f rev.txt
echo  >> rev.txt
```

Other packages have been created (differently) by @jelmd

eugenk:
Ah okay. I noticed that the package `hets-server-core` is updated less frequently than `hets`.
Could we implement a script that updates and rolls out all ubuntu packages at once?

cmaeder:
Yes, "we" could, but please minimize my work. (btw. `hets -update` still works on amd64-wily but `hets-server -update` was never implemented.)

eugenk:
Sure, I don't suggest you to implement the script. I think it's important to know what we need to consider when creating such a script (e.g. packages that need special treatment for their creation). But if it's all in the scripts `utils/debian/auto-package/package_*`, I think we can manage it.

cmaeder:
@sternk created this script initially (and I messed it up, copied it to `package_server` and messed it up further)

cmaeder:
My recent additions of `-dep-suffix ""` to `ghc -M` in `UbuntuMakefile` and `ServerMakefile` will fail for older distros (like trusty up to vivid) that use ghc-7.6! https://ghc.haskell.org/trac/ghc/ticket/9287

https://github.com/spechub/Hets/commit/fb55ef11eafa2d718e848faa3fd45b0e50c18d2e

These files are used in https://github.com/spechub/Hets/blob/master/utils/debian/auto-package/debian-common/rules  (resp. debian-server/rules). Just compiling using `-O` (or `make hets-optimized`) made the ppa build machines fail due to resource exhaustion.

jelmd:
[...]

PR #1708, opened 2017-06-08
Setup Haskell-Stack

eugenk:
[Stack](https://docs.haskellstack.org) is a cross-platform program for developing Haskell projects. It allows us to specify and install the exact versions of the GHC and all Haskell-packages that shall be used in the build process.

The general problem is that different developers use different versions of GHC and Haskell-libraries. This leads to errors where one developer can build a nice and stable Hets-binary while other developers can't even compile it.

Stack solves this problem by reading the specification in the `stack.yaml` and installing an appropriate GHC as well as Haskell-libraries in `~/.stack`. Only these are used when building Hets. This means that you do not need a GHC installed in the system.

As far as I understand, you are currently using system-packages to ensure that you have Haskell-libraries (with `sudo apt-get build-dep hets-desktop`). This step will have no effect on the build process with Stack because it only uses sandboxed libraries.

As of this branch, you only need to [install Stack](https://docs.haskellstack.org/en/stable/install_and_upgrade) (use the generic Linux option if you are on Ubuntu).
Inside the project directory, calling `stack setup` will install GHC and  `stack build --only-dependencies` will install libraries from http://hackage.haskell.org. You don't need to execute these two commands, though, because `make` does it for you.
See the changes in the README.md for required system packages if you encounter problems.

There are several things in this pull request, I need your help with:
* [ ] Check if some parts of the Readme.md can be removed. Especially at the headline "Installing Hets under Ubuntu / Hets development"
* [ ] Check if you can remove build-dependencies from the `hets-desktop` Ubuntu-package.
* [ ] Improve the Makefile/var.mk. I am not very familiar with GNU Make and I only present a working version, but not an elegant version of these two files. In order to get stack to install GHC and the dependencies, it needs to be run *before* all the `ghc-pkg` checks are made. However, they are made with every single call of `make`, because they are used to set variables. So I invoke `stack setup` (install ghc if not installed) and `stack build --only-dependencies` (build and install libraries, if not already installed) to also set variables before the `ghc-pkg` checks. I can't imagine there's no better way to do this.


var.mk
```diff
@@ -33,8 +33,9 @@ STACK_SETUP := $(shell $(STACK) setup > /dev/null)
 STACK_INSTALL_DEPENDENCIES := $(shell $(STACK) build --only-dependencies $(STACK_DEPENDENCIES_FLAGS) > /dev/null)
 
 GHCVERSION := $(shell $(STACK_EXEC) ghc --numeric-version)
-ifneq ($(findstring 7., $(GHCVERSION)),)
-GHC7RTSOPTS := -rtsopts
+GHC_RTSOPTS_AVAILABLE := $(shell expr $(shell $(STACK_EXEC) ghc --numeric-version | cut -d '.' -f 1) \>= 7)
```
jelmd: What does this 'stack exec --' buy here?
Also it is too complicated and clutters the Makefile w/o reason. Using `VAR += $(shell [ _cond_ ] && echo _opts2add_)` is better. 
last but not least since it gets touched anyway: it can be made more efficient and correct: e.g. with `GHCRTSOPTS := $(shell X=$(GHCVERSION) && [ $${X%%.*} -gt 7 ] && echo '-rtsopts')`

.travis.yml
```diff
@@ -1,18 +1,30 @@
...
+  - sudo apt-add-repository -y ppa:hets/hets
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq darwin
 install:
  - sudo apt-get install --no-install-recommends `dpkg-checkbuilddeps debian/control 2>&1 | cut -f3- -d":" | sed -e 's,([^)]*),,g' -e 's,openjdk-.*-jdk,,'` || true
  - sudo apt-get install openjdk-7-jdk
 script:
- - ghc -V
  - make
+ - stack exec -- ghc -V
```
jelmd: wouldn't it make sense to see the version before it starts to build hets?

README.md
```diff
@@ -170,6 +170,16 @@ If you wish to make larger changes we generally recommend [forking](https://help
 
 ### Build Hets using Stack
 * [Install Stack](https://docs.haskellstack.org/en/stable/install_and_upgrade) (use the generic Linux option if you are on Ubuntu).
+* Install build- and GUI-dependencies
+  * Ubuntu:
+    ```
+    sudo apt-get install libglib2.0-dev libcairo2-dev libpango1.0-dev libgtk2.0-dev libglade2-dev libncurses-dev
+    ```
```
jelmd: If they are really required, shouldn't those be added to debian/control's `Build-Depends:` instead of getting out of sync with it? Adding something like `mk-build-deps --install debian/control` might than be ok for flubuntu ;-).

BTW: To find out, which runtime libraries are required, on linux based envs one may use `readelf -d $binary` or otherwise on UNIX based envs `elfdump -d $binary` ...

jelmd:
Wrt. 'setup and build' stack: Your problems come from the fact, that you try to mix up different playgrounds. Strictly spoken, if you do this, then you have to be consequent and build _all_ the required haskell as well as native libs, which in turn means, that you need to build your own OS distro, which includes hets. That's of course not really that smart. Therefore it is usually very helpful to draw a line, where it makes sense. The concern here is to build hets **and not** to build any non-hets-project alias external libs (no matter, whether native or haskell) or even the tool chain required to build it.

So the Makefile should simply check, whether stack is there, if so use it, otherwise use the  traditional way to build what is needed.

Makefile
```diff
@@ -13,14 +13,18 @@ include var.mk
 HETS_VERSION ?= $(shell printf `cat version_nr`)
 
 # We assume ghc 7+
-GHCVERSION := $(shell ghc --numeric-version)
+GHCVERSION := $(shell $(HC) --numeric-version)
```
eugenk: Thanks. I fixed it with 756f31c6d.

Makefile
```diff
@@ -13,14 +13,18 @@ include var.mk
...
+GHC_PARALLEL_FEATURE := $(shell expr $(shell $(HC) --numeric-version | cut -d '.' -f 1) \>= 8)
 NO_BIND_WARNING := -fno-warn-unused-do-bind
 HC_WARN := -Wall -fwarn-tabs \
   -fwarn-unrecognised-pragmas -fno-warn-orphans $(NO_BIND_WARNING)
 # uncomment HC_PROF for profiling (and comment out packages in var.mk)
 # call resulting binary with a final +RTS -p to get a file <binary>.prof
 #HC_PROF := -prof -auto-all -osuf p_o +RTS -K100m -RTS
 HC_OPTS += $(HC_WARN) $(HC_PROF)
+ifeq "$(GHC_PARALLEL_FEATURE)" "1"
+    HC_OPTS += -j
```
eugenk: Thanks for all the input.

The `-j` option *for GHC* is safe: GHC waits for all the dependent files to finish compiling before it starts compiling the target file. `-j` *for make* was not safe with the old Makefile (before you rewrote it) because it did not wait for the dependent targets. I don't know if this is fixed with your new Makefile.

Christian had no chance to include this flag for GHC, because he wasn't involved with Hets anymore when it was introduced with GHC 8.0.

When given only `-j` (instead of, for instance, `-j 2`), it uses all available CPU cores to compile Hets. Each core is assigned a different source file to compile.

Because there were (or still are?) problems with `-j` *for make*, I would rather not couple it with `-j` *for GHC`. Is there another option to let the user decide whether or not to use it? I would like to activate it by default because it significantly reduces compiling time (2 mins vs. 8 mins on my laptop).

Makefile
```diff
@@ -461,7 +465,8 @@ derived_sources += $(drifted_files) Driver/Version.hs $(hs_der_files)
     count fromKif release cgi ghci build-hets callghc \
 	get-programatica check_desktop check_server check_cgi \
 	install install-common install-owl-tools archive \
-	build-indep build-arch build binary-indep binary-arch binary rev.txt
+	build-indep build-arch build binary-indep binary-arch binary rev.txt \
+	stack-setup stack-install-dependencies
```
eugenk: No. That's why I'm asking for help ;)

Oh, and I see that this is something I forgot to remove after I failed to invoke Stack as a dependent target at the right time and then moved to call it for a variable in the var.mk.

README.md
```diff
@@ -167,6 +167,19 @@ If you want to participate in the Hets development feel free to tell us via our
...
+* Build Hets with one of the following:
+  ```
+    make
+    make hets
+    make hets_server
+  ```
+  This uses Stack to build the Hets[-Server] binary.
+  During this process, the specified version of GHC is installed in the user directory, all dependencies are built and finally, the Hets[-Server] binary is compiled.
+
+
```
eugenk: I agree, but that should be done in another issue.

var.mk
```diff
@@ -14,14 +14,29 @@ PROGRAMATICA_SRC_FILE ?= \
...
+else
+    STACK_DEPENDENCIES_FLAGS :=
+endif
+
+# Dummy variables to install GHC and install the Haskell-dependencies.  These
+# must be called before ghc or ghc-pkg is called because the packages need to
+# be installed before they are checked.
+STACK_SETUP := $(shell $(STACK) setup > /dev/null)
+STACK_INSTALL_DEPENDENCIES := $(shell $(STACK) build --only-dependencies $(STACK_DEPENDENCIES_FLAGS) > /dev/null)
+
```
eugenk: That's the point where I need the most help. It must be called before all the ghc-pkg calls are made (which are in the var.mk to setup, for instance, `HAXMLVERSION`) 
and I don't know how to do this. As you can see, I am not at all an expert of Make.

.travis.yml
eugenk: Agree. Then, I need to call `stack setup` before the `make`. Done with 756f31c6d.

README.md
eugenk: I did not know about this file. Thanks. I changed it with 96453121b.

eugenk:
>  build all the required haskell as well as native libs

This is exactly what Stack is for.

> which in turn means, that you need to build your own OS distro, which includes hets

I think you exaggerate a lot here.


The point is to have the same libs on the same versions for all developers regardless of their OS/distribution. The previous workflow only works well if *all* Hets developers use Ubuntu and agree on the same GHC version (What if they have other Haskell projects with other requirements?). Well, that's not the case. We have three developers on two completely different OSes with three different GHC versions.

Setting up the environment (installing the GHC and Haskell-libs) that is required for the project is IMHO part of the build process. At least installing the libs is done with other languages, too (Rust, JavaScript, Ruby to name a few). Installing the right compiler version is not very different from installing the libs in this case.

---

I agree, that I mix two playgrounds, but they are rather a) using make targets to get things done and b) using make variables to get things done.

A solution would be to run a make target before all the variables like `HAXMLVERSION` are set. Is that in any way possible? I don't say that these must stay variables. If no such fully automatic approach can be done, we should create `make stack` to initialise Stack, as you suggested.

README.md
jelmd: Ok

var.mk
jelmd: Not sure, what it really does. If it is really only for setting up/building stack, than a simple target for convenience like this would be IMHO sufficient:
```
stack:
    $(STACK) setup
    $(STACK) build --only-dependencies $(STACK_DEPENDENCIES_FLAGS)
```

So if the developer hasn't set it up yet, he just needs to 'make stack' once and than forget it. As said, building hets should NOT be interweaved with setting up the tool chain, that's it.
However, because you made the setup part of the Makefile, you also need to include a receipt to remove the so installed stuff cleanly, i.e. w/o removing anything, which is not hets related/shared by other build stuff. So you hopefully see, that pulling in the stack setup is not very wise and opens a plethora of other problems - just a waste of time to fix it. The better thing one would do is to document 'stack' as an [optional] requirement and possibly add some links, which point to articles, how to setup this stuff.

README.md
jelmd: Yepp, but please check, whether they are really needed! If they are just needed by stack to rebuild external libs/modules, they should not be included there, but rather in stack's env definition/setup files ...

var.mk
eugenk: > Not sure, what it really does

Stack is for Haskell what Bundler is for Ruby. `stack setup` and `stack build --only-dependencies` together are like `bundle install`. The first command installs GHC and the standard libs (specified by the key `resolver` in the stack.yaml) to `~/.stack/` and the second command installs extra-dependencies (specific to the project; not included in the standard libs; specified by the key `extra-deps`) to `./.stack-work/`.

It's no problem to remove `./.stack-work` again with a `clean_stack` target.

README.md
eugenk: They are all needed. I checked with a fresh Ubuntu 16.04, Ubuntu 17.04 and a fresh Ubuntu Server 14.04. Otherwise Hets won't build.

var.mk
eugenk: Oh, and Stack should *not* be optional. There's no other feasible way to manage compiler- and library-versions with all developers.

README.md
jelmd: Just wondered, because even the current hets-desktop doesn't have a runtime dependency to ncurses and the environment, where I usually build ubuntu packages does not have the libncurses-dev installed either ... Anyway, could be an indirect dep ...

eugenk: It depends on the Haskell package `terminfo` (which is strictly speaking a dependency of a dependency) which needs libncurses.

jelmd: OK, so should not go into the control file.

eugenk: It should not? If libncurses is not there, Hets's dependencies (and thus, Hets) won't build.

var.mk
jelmd: There are many more ;-)

Anyway, I'm glad, that this isn't a C/C++ or Java-Project ... ;-)

README.md
jelmd: Just checked: The script I use to setup the env and build it is:
```
...
function doPrepare {
    DEBPKGS='gnupg2 git ant openjdk-7-jdk  ghc libghc-haxml-dev libghc-tar-dev libghc-glade-dev libghc-http-dev libghc-warp-dev libghc-wai-extra-dev libghc-fgl-dev libghc-xml-dev texlive-latex-base texlive-latex-extra texlive-fonts-recommended'
    (( ${LSB_RELEASE//.} < 1510 )) && DEBPKGS+=' libghc-haskeline-dev'
    # from ppa.launchpad.net/hets
    DEBPKGS+=' libghc-aterm-dev libghc-hexpat-dev libghc-programatica-dev libghc-uni-htk-dev libghc-uni-graphs-dev libghc-uni-udrawgraph-dev'
    # zwecks tests
    DEBPKGS+=' darwin spass'    # NOTE: have indirect dep on eprove
    # für HTML API doc
    #DEBPKGS+=' libghc-haddock-dev libghc-missingh-dev'

    prepareRoot4build

    ${TAR} xJf ${SRC}/${HETS}.tar.xz
    cd ${TMPBUILD} || exit 1

    for p in ~(N)${SDIR}/patches/*.patch ; do
        ${PATCH} -p1 -E -b -z .orig -i $p
    done
    # avoid not yet ready meta-packages
    sed -i.orig -e '/TBD/,$ d' debian/control
}

function doMake {
    cd ${TMPBUILD} || exit 1

    checkMakeVars
    Log.info "make log goes to ${MAKELOG}"
    rm -rf debian/root ../hets-{common,desktop,server}_${RELEASE}_*.deb
    fakeroot make binary 2>&1 1>${MAKELOG} |tee /tmp/stderr.log

    [[ -d tmp ]] || mkdir tmp
    mv ../hets-{common,desktop,server}_${RELEASE}_*.deb tmp/
}

function doPackage {
...
```

and I guess, I see the reason: IIRC haskeline requires terminfo in ubuntu < 15.10, only ...

var.mk
eugenk: > There are many more ;-)

Please elaborate :)

README.md
eugenk: The code/documentation from haskeline says:
```
-- There are three main advantages to the terminfo backend over the portable,
-- "dumb" alternative.  First, it enables more efficient control sequences
-- when redrawing the input.  Second, and more importantly, it enables us
-- to draw on multiple lines, so we can wrap long input strings.  And third,
-- the backend adds some extra key sequences such as forwards delete.
--
-- (The "dumb" terminal also allows editing of long input strings, but is
-- restricted to only one line and thus only shows part of the input at once.)
```

eugenk: I have no idea what it is used for in Hets and I don't want to decide whether or not to include this (and thus, require libncurses-dev). @tillmo, can you please give us some input on this topic?

jelmd: Well, I probably took it out, because the more recent releases don't have this package anymore - it is now part of the ghc package and thus it is "automagically" there. However, because you bypass the system, it hits you right into the face ...

BTW: Take care about what you add to the debian/control file. If it is not sufficient or invalid, this will break launchpad builds - and dealing with launchpad errors is horror at its best/costs a lot of time! So make sure, that at least for the main target (current LTS release) all specified packages actually exist and are sufficient to build the package! launchpad is an extremely picky ...

Makefile
```diff
@@ -580,16 +580,16 @@ pretty/LaTeX_maps.hs: utils/words.pl utils/genItCorrections \
...
 
-### removes all *.o, *.hi and *.p_o files in all subdirectories
+### removes all *.o, *.hi and *.p_o files in all subdirectories except for .stack-work
 o_clean:
-	@find . \( -path ./.stack-work -prune \
-		-name '*.o' -o -name '*.hi' -o -name '*.p_o' \
-        -o -name '*.dyn_hi' -o -name '*.dyn_o' \) -exec rm -f {} +
+	@find . \( -name '*.o' -o -name '*.hi' -o -name '*.p_o' \
+			   -o -name '*.dyn_hi' -o -name '*.dyn_o' \) \
+			-o -path ./.stack-work -prune -type f -exec rm -f {} +
```
jelmd: Ehmm, no - I doubt that it does, what it should ;-) - shortcut rules are in effect! I.e. if for `(A) || B` A evals to true, B gets not evaluated. So actually the expression above just considers everything not matching the patterns enclosed in () as rm candidate (subject for B). Unfortunately (if -path _pattern_ works as you say), the only candidate for -exec expression becomes the file ./.stack-work, so IMHO under normal conditions the find expressions always evaluates to false/doesn't remove anything.

README.md
tillmo: @eugenk, I think this is used for the Hets interactive interface `hets -I`.

Makefile
```diff
@@ -615,7 +622,7 @@ realclean: clean java_clean
 	@$(RM) -f *.bin debian/changelog*
 
 ### additionally removes generated files not in the repository tree
-distclean: realclean clean_genRules
+distclean: clean_stack realclean clean_genRules
```
jelmd: Actually distclean should remove everything, what got produced during one or more runs, i.e. rollback in a way that it appears as make has never been run. So I guess, there needs to be done more than just rm stack's work dir?

eugenk: It removes everything that got produced in the project's directory. 

How about this: 24944b84c

Makefile
eugenk: Fixed in fa6878c47.

README.md
eugenk: Do we need/want to have the [quoted improvements of `terminfo`](https://github.com/spechub/Hets/pull/1708#discussion_r121206506) for the interactive interface? If we don't, we can remove the dependency of `libncurses-dev`.

If I remember correctly, the interactive interface may draw multiple lines, so removing `terminfo` could be a step back.

From the perspective of a user, I would like to have the "forwards delete" feature. So it's a plus for `terminfo`.

tillmo: yes, I think we should keep it.

Makefile
jelmd:
```
make distclean                                make: stack: Command not found
/bin/sh: 1: [: -ge: unexpected operator
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
make: stack: Command not found
/bin/sh: 1: [: -ge: unexpected operator
```
As said: please make sure, that stack is there before using it blindly!

Also wrt. all other stack related changes make sure, that it can still be build via launchpad (unless the PPA is not needed anymore)! Not sure, but IIRC, downloading something during the build is not allowed. You could do so by setting up your own ppa and try to build a simple example package using stack. When you have worked out, how this needs to be done, it needs to be "merged" into hets build process ...

eugenk: Done the check with 96a5d8fc6

About launchpad: Sorry, I'm not that familiar with Debian/Ubuntu and don't really understand what you say about Launchpad and PPAs.

Makefile
```diff
@@ -586,9 +586,9 @@ clean_genRules:
 
 ### removes all *.o, *.hi and *.p_o files in all subdirectories except for .stack-work
 o_clean:
-	@find . \( -name '*.o' -o -name '*.hi' -o -name '*.p_o' \
-			   -o -name '*.dyn_hi' -o -name '*.dyn_o' \) \
-			-o -path ./.stack-work -prune -type f -exec rm -f {} +
+	@find . \( \( -name '*.o' -o -name '*.hi' -o -name '*.p_o' \
+			      -o -name '*.dyn_hi' -o -name '*.dyn_o' \) \
+			   -o -path ./.stack-work -prune -type f \) -exec rm -f {} +
```
eugenk: The intention is already in the comment right above the target:
> removes all *.o, *.hi and *.p_o **files** in all subdirectories **except for .stack-work**

Done the simplification in 91432c9a2

var.mk
```diff
@@ -1,5 +1,10 @@
 # to be included by Makefile
 
+STACK_EXISTS := $(shell command -v stack 2> /dev/null)
+ifndef STACK_EXISTS
+    $(error "stack is not available please install haskell-stack")
+endif
+
```
jelmd: There is no need, that stack exists, neither for *clean targets nor for the other stuff! As said, setting 'STACK=/bin/true' and `STACK_EXEC=`, i.e. to an empty string should be sufficient.

Makefile
```diff
@@ -584,11 +584,12 @@ clean_stack:
...
 	@$(RM) $(generated_rule_files) $(gendrifted_files) $(hs_clean_files)
 
-### removes all *.o, *.hi and *.p_o files in all subdirectories except for .stack-work
+### removes all *.o, *.hi and *.p_o files in all subdirectories except for
+### .stack-work, where the compiled dependencies reside
 o_clean:
-	@find . \( \( -name '*.o' -o -name '*.hi' -o -name '*.p_o' \
-			      -o -name '*.dyn_hi' -o -name '*.dyn_o' \) \
-			   -o -path ./.stack-work -prune -type f \) -exec rm -f {} +
+	@find . -path ./.stack-work -prune -type f -o \
```
jelmd: the '-o' should be at the start of the line, not at the end

README.md
jelmd: Ehmm, you are mixing up two different things, i.e. the "new build" which bypasses the system, and the proper build which relies on the system. It is not a question, whether to drop haskeline (terminfo), but how they are build and used by ghc when it finally links the binary. If you bypass the system, there is no haskeline and it probably gets build with given defaults, which might be with libncurses directly or indirectly via terminfo, which than requires the additional dev packages. If one does not bypass the system, the haskeline comming with the ghc package gets used: this is not linked or required to link against libncurses or uses a terminfo, which itself does not require libncurses (or may require it, but than this is solved wrt. the terminfo module and its given package dependencies). So it shouldn't surprise one, that for different ways of building stuff there are different dependencies .... Wrt. ubuntu packages, those have to go into debian/control, with the hybrid/system bypassing build it probably needs to be persisted  somewhere else ...

Makefile
jelmd: Oh, what a lucky man :). Unfortunately because you are changing the build system, you need to suck this bitter pill and make sure, that it works on launchpad as well. PPA == private package archiv. hets PPA is used so that ubuntu users just can do an `apt-get install hets`, etc. to install hets & friends. For this one needs to create a debian source package description and upload this incl. the sources to launchpad. It than automatically builds the binary packages and finally publishes them to the public PPA mirrors. If there is an error in the source description package or in the build itself, you need to do all the stuff again (no shortcuts). For details see: https://help.launchpad.net/Packaging

var.mk
eugenk: I don't understand it.

> As said: please make sure, that stack is there before using it blindly!

Here you say that stack needs to be there.

> There is no need, that stack exists

Here you say, it doesn't need to be there.

I understand that it doesn't need to be available for clean targets etc, but the `var.mk` uses it a lot because the "what version of package x is installed" checks *need* to use stack. And these are done for *every* target. So actually, stack needs to be there for clean targets.

jelmd: Close to: When you wanna use stack as currently done in the Makefile, you need to make sure, that it is actually there. If it is not there, don't use it - and this can be done by defining the related vars as shown above.

eugenk: But stack should be mandatory because we need to guarantee a specific build environment (GHC changes way too much to ignore this).

Makefile
eugenk: If downloading is not allowed during launchpad building, then there seems to be no way to ensure this environment for the PPA... Even if we don't use Stack for the PPA and if manage to use the same GHC as in development *consistently*, we are stuck with the Debian-packaged Haskell-packages (e.g. glade) and cannot use the versions specified by the Hets.cabal and stack.yaml. This may lead to differences between the developed Hets and the Hets that users install via apt-get. This pull request is supposed to remove this difference.

tillmo: Can't we auto-generate some debian package from the Haskell packages installed by Stack? And then use this debian package for the PPA build process? We would need a flag that indicates whether we are in the PPA build process or not.

tillmo: By the way, in principle, debian packages can be auto-generated. I have already done this.

eugenk: I'm still struggling to understand the debian build system/launchpad. I find the build system very confusing, but maybe it's only because of the docs.

jelmd: Yepp, docs are very shallow and often really vague - they probably embraced the "try and error" * too much ;-)

IIRC, Christian used pbuilder, debuild and possibly other stuff in utils/debian/ (which is dead meat since last major build system update 1+ year ago), which is very inefficient, introduces a lot of overhead (ein vermotteter Flickenteppich aus paranoidem/kranken Zeux ;-) )  and steals one a lot of time ...

Therefore I decided to use source format 2.0 and to create the required description as well as changes file incl. src archive for each release directly by myself using the "lower level" tools directly (dpkg-genchanges, dpkg-source + sha1, sha256, md5, gpg2, tar, sftp) in one row.
But perhaps you find a different way to do it with the deb[u]il[d] tool.

Wrt. download or not: as said, I'm not sure, maybe this have changed. At least it is not documented, so just try if you need it.

Wrt. "all have to use the same": well, good software usually uses macros/defines or similar things to honor the users environment/wishes and to work around platform/release specific issues, like var.mk does. When you think, that stack solves your problems magically, then I guess you'll hit sooner or later a wall, which requires to think over it again ...

BTW: I consider all the "I re-invent the wheel and do my own packaging stuff/bypassing the system in use" very dangerous - for development it is ok wrt. being able to work on the cutting edge, but the distributed stuff, which people aka end users will actually use on their systems should smoothly integrate into this environment and use system packages/software as much as it can.

eugenk:
> Not sure, but IIRC, downloading something during the build is not allowed.

It's not possible. Search for "Downloading lts-8.17 build plan ..." in https://launchpadlibrarian.net/324942262/buildlog_ubuntu-xenial-amd64.hets-src_0.99_BUILDING.txt.gz
There is no network connection for the build bot.

So I guess we would need to package haskell-stack (the version in the repositories is only "sort of up to date" for 16.10 and greater while the version for 16.04 is way outdated.) or drop support for 16.04 (which I assume is not ok because it's the latest LTS release).

Additionally, we would need to package the stack build plan used in the stack.yaml (`lts-8.17`) which consists of GHC and lots of haskell packages. Next, we would need to package all haskell-packages from the extra-deps.

Otherwise we can't guarantee that the version in the PPA is the same as in our development environment and reproducibility is gone.

tillmo:
can the necessary packaging of the Haskell deps be automated?

eugenk:
If it's possible to automate packaging at all, it should be possible to automate it in this case as well.

However, there are a few things to do, some of which (I think) will not allow Hets to make the jump to the `universe` repository:

* We need to package the current version of Haskell-stack. There is no package for Ubuntu 14.04 LTS, the package for 16.04 LTS is severely outdated and useless and the package for the newer Ubuntu releases are slightly outdated and may be useful, but I'm not sure.
* If we somehow manage to package stack along with a build plan, then we need to install Haskell-packages to `~/.stack/`. This is not very hard to do because it seems to suffice to delete the compiled files from the development machine and package the rest as source files because Stack keeps the source files.
* We need to install GHC to `~/.stack/` which is difficult without internet access. Stack downloads a preconfigured version of GHC, compiles it and installs it inside this directory. Even though stack keeps the downloaded tar file with the GHC-sources, it does not suffice to delete the compiled GHC. Stack still wants to download the sources from the internet.
* We need to install sources of the `extra-deps` to `Hets/.stack-work/` or alternatively include them in the `hets-src` package.

eugenk:
My plan is to create these packages in a PPA:
* `haskell-stack` which is just the plain Haskell-stack. Nothing added
* `hets-src-builddep-stack` which adds all the build dependencies for Hets to `~/.stack`. The source package contains the whole directory structure of if *after* `stack setup` has been run, with the exception that the compiled files are deleted again. In a build step, the binaries are compiled again, but don't need to be downloaded anymore. I need to figure out how to do this with GHC.
It [build-]depends on `haskell-stack`.
* `hets-src` (exists already) which build-depends on `hets-src-builddep-stack` and includes everything it already has *plus* the `.stack-work` directory after running `stack install --only-dependencies` and deleting the binaries from it.

All this just to have an Ubuntu package that is guaranteed to be the same as we create in the development environment.

tillmo:
sounds good. For the requirement that ghc needs to be in `~/.stack`, couldn't you rely on ghc being installed with a ghc package and then just create a symlink in `~/.stack`?

eugenk:
Unfortunately not, because there is no current GHC package for the older Ubuntu versions (I could not find a PPA either).

eugenk:
Turns out it's much more involved. There are no libghc-* packages in the right versions for many libs (in `artful` repositories):
```
Configuring stack-1.4.0...
hlibrary.setup: Encountered missing dependencies:
aeson >=1.0 && <1.2,
annotated-wl-pprint -any,
clock >=0.7.2,
cryptonite-conduit -any,
file-embed >=0.0.10,
hackage-security -any,
hpack >=0.17.0 && <0.18,
http-client >=0.5.3.3,
http-client-tls >=0.3.4,
http-conduit >=2.2.3,
lifted-async -any,
microlens-mtl -any,
mono-traversable >=0.10.2 && <1.1,
neat-interpolation >=0.3,
optparse-applicative ==0.13.*,
pid1 ==0.1.*,
safe-exceptions -any,
smallcheck -any,
store >=0.2.1.0,
text-metrics >=0.1 && <0.3,
th-reify-many -any,
unicode-transforms >=0.1 && <0.4
```
Most of the libghc-* packages are **way** outdated. For example: latest aeson lib is 1.2.1.0 while the latest deb package of it is libghc-aeson-dev 0.11.3.0-2. [See here for the list of versions in between.](http://hackage.haskell.org/package/aeson) I wonder how the build system can work at all, if these packages don't get updated...

eugenk:
Even if we manage to build Stack in a PPA for artful, we would need to build GHC8 (and lots of libghc-* packages) for previous releases of Ubuntu since this is a dependency of the current Stack... I think this is not feasible.

Another approach would be to bundle a local Stack plus dependencies with the `hets-src` package and call this local Stack to build the package. I'm not quite sure if I could do that or if I'm even allowed to do it for licensing reasons.

I don't like the Debian build system at all...

tillmo:
Nevertheless, the provision of Ubuntu packages has greatly increased the visibility and usage of Hets. Only few users will build Hets from sources. Would it be easier to concentrate just on Ubuntu 16.04 LTS, and then forthcoming LTS versions (18.04, 20.04 etc.)?

eugenk:
It does not really matter which Ubuntu version we concentrate on because even the next one, 17.10, does not have up to date libghc-* packages. It's just the first one that has an up to date GHC.

In total we need to supply 
* system-level GHC 8.0
* every libghc-* depedency of Stack (there are 141 of them) - well, at least the ones that are too old to be compatible
* Stack itself
* GHC inside of Stack

[Apparently snap packages can be built with internet access](https://lists.ubuntu.com/archives/snapcraft/2016-December/001950.html) which would make this so much simpler.

If we can't create snap packages of Hets, we should take a closer look at how stack could be [bundled](https://answers.launchpad.net/launchpad/+question/239751) with the hets-src package.

jelmd:
I agree with eugenk. IMHO this burns/has already burned down a lot of valuable energy/resources for almost nothing.

I suggest to make it like many other OSS does: 
1) There is one "stable" release, which gets distributed in related distro packages and aligns with the distro, i.e. use what is already there and not stack/do not try to rebuild/dup/reship the "whole" distro.
2) And there is a "cutting edge" aka developer version, which could be build by the interested developer itself using stack. For convenience one of those developers or bot can build packages of it for main distros/releases and put it into a dedicated download area, where people can download and install it locally. (BTW: Wasn't the plan to do it like this for the mac packages anyway?)

Last but not least I'm a little bit surprised about the "**way outdated**" usage in this context. Just because there is a major does not necessarily mean, the previous version is obsolete/should not be used. Actually I've seen it in a lot of OSS, that one can't simply recompile it using the latest version of all libs/components, because the SW relies on some features, which are not available in the new version or makes it simply to fat, because the lib/etc. now contains a lot of other useless bloat ... So using old SW is not unusual and sometimes even intended. IIRC aeson is such an example: many stuff uses still the stable 0.x version, because it is horror to adjust it to the "new"/still buggy/unstable 1.x version, which is completely different ...

cmaeder:
I've no idea about stack but used to use https://launchpad.net/~hvr/+archive/ubuntu/ghc plus its cabal to install the remaining packages.

eugenk:
> this burns/has already burned down a lot of valuable energy/resources

Yes. Still, what I would like to look into is making a snap package of Hets. The referenced email suggests that one has internet access during the build phase and snap packages even run on different distributions. Maybe we could do all of these things

1. Keep the traditional build system for a PPA with the old libs and so on, but not for every version of Ubuntu that is still supported by Canonical. 
2. Host a deb package that is built with Stack. I could build and upload them when I build and upload the packages for Arch Linux and macOS
3. Build a snap package for any Ubuntu from 16.04 upwards and other systems (not necessarily an issue of this PR) which we recommend to use.

> For convenience one of those developers or bot can build packages of it for main distros/releases and put it into a dedicated download area

So, host a simple .deb file for amd64?

> BTW: Wasn't the plan to do it like this for the mac packages anyway?

Yes and no. The de facto standard package manager for macOS is [homebrew](https://brew.sh) and I supply the Hets package there. It either downloads the binary package that is hosted somewhere or (if there's something wrong with the binary, e.g. checksum mismatch, the server is offline, etc) it builds Hets locally.

> I've seen it in a lot of OSS, that one can't simply recompile it using the latest version of all libs/components, because the SW relies on some features, which are not available in the new version

True, but if it really depends on the newer version and there is no package for the newer version, we have a problem (like in the example of Stack). Also, it may become a safety/security concern to stay on older versions of libs (they won't receive security updaters for ever). This definitely is a concern for Hets because of the server mode.

eugenk:
> I've no idea about stack but used to use https://launchpad.net/~hvr/+archive/ubuntu/ghc plus its cabal to install the remaining packages.

Cabal needs internet access. How did you do this for the PPA?

cmaeder:
I only used the hvr repository for testing with travis. For the hets ppa I relied on the platform ghc and only installed a few additional haskell packages in our hets ppa.

jelmd:
@eugenk
- yes wrt. 1. , 2. and .deb file
- wrt. 3. - snap packages: Is this another Ubuntu project like upstart and unity? I wouldn't waste any time on it. May be when it is widely used and known to be a good thing, but this will take some years ...
- wrt. homebrew: aha, thought it gets always build on localhost like the port* stuff.
- wrt. security: Not sure, but IMHO most haskell packages provide security just by obscurity. I.e. I guess most packages just focus on features and not on fixing any security related thing, which means, its a situation like the openssl one before the heartbleed problem got public - everyone assumes, that the makers of a module know, what they do and have thought through any possible corner case ... or thinks, that "the language" per se produces only secure stuff (BTW: Wondering, whether any haskell module related security vulnerabilities have been published in the past ...). Therefore I trust/rely on system packages slightly more than the ones from a certain "junk yard" because the related SW gets at least some love from the vendor et al. and gets adjusted, so that every distro provided package can be build with the tool chain and modules provided by the system/distro.
- wrt. "need new feature": the usual action is to drop support for releases, which do not provide the required stuff, or to backport it if really needed, or just simply restrict itself to the least common multiple wrt. availability, or to write some code, which does what is needed (possibly in a more efficient way) ...
- wrt. cabal: If you use the current Makefile, which can be used as DEBIAN/rule file as is, there is no need for cabal.

eugenk:
wrt 3: I took a look at snap and it's supported by some of the big distributions as well (see https://snapcraft.io/docs/core/install) but it has some severe limitations wrt. to file system access. So, snap is out of the picture.

tillmo:
Since you wrote so negatively about the Debian package system, let me try to explain some philosophy behind it. Debian (together with its derivatives) is the package management system that provides the largest number of free software packages (several times more than rpm-based systems, and an order of magnitude more than e.g. arch linux). The restriction that you cannot download sources during the build process is motivated in several ways:
* it can be (and is) used to ensure that packages are really open source,
* it enforces a discipline that package dependencies are tracked correctly and one and the same package is not imported through multiple ways,
* and via this, it ensures modularity and scalability of the package system.

eugenk:
I changed this such that stack is not used if it's not available. I also reverted the debian/control file. This now disables reproducibility for apt-packages of Hets.

var.mk
```diff
@@ -1,5 +1,27 @@
 # to be included by Makefile
```
jelmd: Because of readability/flexibility I prefer:
```
STACK ?= $(shell command -v stack > /dev/null)
STACK_EXEC :=
STACK_TARGET :=
STACK_UPGRADE_TARGET :=
STACK_DEPENDENCIES_FLAGS :=
ifneq ($(STACK),)
    STACK_EXEC := $(STACK) exec --
    ...
endif
```

Last but not least `printf` should be used instead of `echo` - there are many different echo implementations/shell builtins and it emits per default a redundant linefeed at the EOL, which may cause add. headaches from time to time ...

var.mk
```diff
@@ -14,14 +36,19 @@ PROGRAMATICA_SRC_FILE ?= \
...
-ifneq ($(findstring 7., $(GHCVERSION)),)
-GHC7RTSOPTS := -rtsopts
-endif
-
 OSNAME := $(shell uname -s)
 OSVERS := $(shell uname -v 2>/dev/null)
 
+ifeq "$(OSNAME)" "Darwin"
+    STACK_DEPENDENCIES_FLAGS := --flag gtk:have-quartz-gtk
+else
```
jelmd: delete else branch, since already initialized in the suggested change above. Also it would be better to move OSNAME, OSVERS setting lines to the top of the file and nest setting it into the  `ifneq ($(STACK),)` body.

eugenk:
I changed it. I tried this on a ppa, but the dependencies cannot be met. See the build log at
https://launchpad.net/~eugenk/+archive/ubuntu/testing/+packages

It looks similar to the problem in the discussion at #1556.

jelmd:
Hmmm, I guess the problem is, that the mentioned packages are neither in the main nor in your own ppa.
Not sure, if there is a way to specify add. PPAs to use, when preparing the build env or at least to "symlink" them ...

eugenk:
Okay, I have added the official Hets ppa as a dependency, but I get this error on xenial and older (but not locally on xenial):

> /usr/bin/make hets-opt
make[1]: Entering directory '/<<PKGBUILDDIR>>'
Makefile:482: recipe for target 'check_desktop' failed
make[1]: [check_desktop] Error 1 (ignored)
utils/genRules -r ShATermConvertible -i ATerm.Lib -o ATC/Id.der.hs Common/Id.hs
utils/genRules: 4: utils/genRules: Syntax error: word unexpected (expecting ")")
Makefile:209: recipe for target 'ATC/Id.der.hs' failed
make[1]: *** [ATC/Id.der.hs] Error 2
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
debian/rules:913: recipe for target 'hets.bin' failed
make: *** [hets.bin] Error 2
dpkg-buildpackage: error: debian/rules build gave error exit status 2

Zesty has the same dependency problem as I had before.

jelmd:
Well, you could compare it with the latest successful builds via  [hets ppa](https://launchpad.net/~hets/+archive/ubuntu/hets/+packages?field.name_filter=hets-src&field.status_filter=published&field.series_filter=) - e.g.  [latest xenial 64bit build](https://launchpadlibrarian.net/278978611/buildlog_ubuntu-xenial-amd64.hets-src_0.99.14-1.16.4_BUILDING.txt.gz).

There one may things being odd. E.g. from the `jars` target usually emits some warnings. Also I don't see a `/usr/bin/make hets-opt`, which suggest, that the Makefile aka rules got screwed up (or perhaps that a depend target failed and thus `hets-opt` gets not made ...

Have you tried to build it locally without stack? `make -n binary` may help to see differences w/o actually building the stuff ...

BTW: no[t all] haskell-* pkgs from hets PPA are available for post xenial releases AFAICS. So builds for artful, zesty will definitely fail.

Last but not least: Take care of strange PPA versioning rules. They are pretty picky and many things simply do not make sense, but trying to take a "shortcut" usually cause also really strange behavior/failed builds ...

jelmd:
Seems, that your last ppa build succeeded. So from my POV (i.e. non-stack), when the debian/control file got restored to master version and latest Makefile changes (tests) reversed, I think it is ok to merge.

eugenk:
It succeeded, but there were still some debug traces in the makefile. I removed them and wait for it to build on the PPA.

jelmd:
Yepp, this is, what I meant with 'reverse'. BTW: You should use the `archive` make target to produce the arc to upload to the ppa - ~20MB is far less then ~1.4 GB ;-)

eugenk:
Is there a script available that you use to create the ppa-package?

eugenk:
Now, it's successfully built without the debug traces.

jelmd:
Script for this horror - of course: see @IKS `~elkner/tmp/lnx/build/develop/hets/Build.sh` - should be readable by everyone (is the outchecked repo I use to work wrt. Linux pkgs ...). The script relies on the repo structure/supporting scripts, so probably not directly aka  standalone/ad hoc useable. However, should give a clue, how I usually do it, which shortcuts may work. All Build.sh have basically 3 stages: doPrepare(), doMake() and finally doPackage() ...

eugenk:
Thanks, I will take a look at it to understand things better.

Is this pull request ready to merge or are there things I should change? (Github says it's not approved yet)

jelmd:
AFAICS there is still the `check_genrules` stuff and debian/control is still not set back to the master version. Just do a `git diff master Makefile` and `git diff master debian/control` to see the final  changes, which would be merged.

eugenk:
Right, I seem to have forgotten to push. The two latest commits are removed now. However, we need `git` and `latexmk` in the debian/control file. Building failed in the PPA without these.

eugenk:
I will attempt another build without these two dependencies, but last time it failed because of them.

PR #1715, opened 2017-08-08
1714 update het casl to dol

tillmo:


mcodescu:
The changes are fine. The only thing that I believe should be done is to document somewhere what is supported in HetCASL and not in DOL, e.g. extensions with arbitrarily complex OMS, or imports of the form "from ... get ...". Otherwise :+1:

tillmo:
OK, but this would belong to the user guide. I will file a separate issue.

PR #1716, opened 2017-08-09
Introduce semantic versioning

### LINKED ISSUE #1713: Switch to semantic versioning
We should switch to [semantic versioning](http://semver.org/) because Hets is not only used as an application but also like a library (Hets-server). In the latter case, it is especially important to know whether or not there were breaking changes.

I suggest to
* Rename the file `version_nr` to `version` because the current file name is a mix of English and German and calling it `version` suffices.
* Release often: Every time a new feature is finished (may be one pull request, may be several pull requests), we should increment the version number in the final pull request of a feature and create a release tag as soon as it is merged into master.
* Add command line options/change their output for version queries:
  * `--numeric-version` only prints `0.100.0`
  * `--version` prints `The Heterogeneous Tool Set, version 0.100.0`
  * `--extended-version` prints `The Heterogeneous Tool Set, version 0.100.0, revision abcdef` which might be useful for bug reports.

Before I attempt to change the code accordingly I want to ask @jelmd:
Should we keep the `rev.txt` with the time stamp? It is also used to generate the file name in the `archive` target. Do we need the time stamp in the archive name if we switch to semantic versioning? Instead of the time stamp, I would like to put the commit hash in that file for the `--extended-version` call (and rename the file to `git_revision`).

eugenk:
This shall fix #1713.


jelmd:
LGTM (wrt. non-haskell part)

eugenk:
Thanks for the changes in the Makefile.

tillmo:
:+1:

PR #1729, opened 2017-08-21
Remove dependencies: wget, cat, tac, grep/ggrep

eugenk:
This removes wget as a dependency. This now uses the package `http-client` to make HTTP calls. It also adds an option to disable TLS certificate verification.

@jelmd, Please tell me if the wget dependency should stay in the Debian control files.

### DIFF
--- a/Common/Http.hs
+++ b/Common/Http.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedStrings #-}
 {- |
 Module      :  ./Common/Http.hs
 Description :  wrapper for simple http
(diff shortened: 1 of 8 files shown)


mcodescu:
The code looks sound to me.

jelmd:
LGTM (non-haskell part).

BTW: If you wanna cleanup dependencies to `cat` and `tac` as well as using POSIX compatible egrep instead of incompatible `grep -e`, have a look at http://iks.cs.ovgu.de/~elkner/tmp/hets-extdeps.patch ;-)

eugenk:
Good point. I added two commits that do this.

eugenk:
Are there any news on the review? Do I need to change things?

tillmo:
it seems that travis fails because http://ppa.launchpad.net/ is down, see also https://twitter.com/launchpadstatus?lang=en

eugenk:
Fixed with fc894d9.

eugenk:
I needed to rebase on master. There were no conflicts, so the diff is the same as before.

tillmo:
actually, I have enabled a rather lax regime: there is no enforcement that new commits must be approved again. Is this useful?

eugenk:
That's absolutely fine for most cases. I just wanted to make sure that you know what the last push is about.

PR #1728, opened 2017-08-21
fixed #1727

tillmo:
renamed predicate Thing:Thing into ThingPred:Thing in order to avoid clashes in Ontohub 1.0

PR #1730, opened 2017-08-21
1722 integrate more tests

### LINKED ISSUE #1722: integrate more tests
in git@github.com:spechub/Hets-lib.git, I have added a script check-files that analyses 259 files (and some more directly and indirectly imported files) of the Hets-lib repository. This should be added to the Hets test suite.

eugenk:
This shall fix #1722.

A test being red shows the proper behaviour of Hets. There is actually a syntax error in Hets-lib/CommonLogic/fUML.clif: CommonLogic comments are indicated by `(cl-comment ...)` and not by `/* ... */`.


tillmo:
it seems that #1728 causes a test to fail. This again is not a problem of the tests, but of #1728, which should not have been approved. I suggest that we revert #1728 and keep it on a branch `ontohub-1.0`, since it is only needed for fixing some Ontohub 1.0 related problem that will disappear in Ontohub 2.0.

mcodescu:
I agree with Till. Sorry, I haven't looked whether the checks have passed and only tested the expected behavior. If I understand the failed test correctly, we could make things work by renaming the predicate Thing to ThingPred in the target of the view, or add an explicit mapping of ThingPred to Thing. But it makes sense not to modify the test for a change that is just a workaround.

tillmo:
The problem was that the tests are introduced only with this pull request - so they were not available for #1728 (at least not automatically).

tillmo:
see #1734

tillmo:
@eugenk: rebasing on master should solve the problem now.

eugenk:
Done

PR #1741, opened 2017-09-06
Split travis build into 'Hets-lib' and 'make check' jobs.

eugenk:
This splits the Travis-CI build into two separate jobs that are executed in parallel. 

If wanted, we can supply *n* different `test/hets-lib-testfiles_i` files that partition Hets-lib and analyse everything. This would result in *n+1* jobs in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,10 @@ before_install:
   - travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
 install:
   - sudo apt-get install --no-install-recommends `dpkg-checkbuilddeps debian/control 2>&1 | cut -f3- -d":" | sed -e 's,([^)]*),,g' -e 's,openjdk-.*-jdk,,'` || true
+matrix:
+  include:
+    - env: WITH_MAKE_CHECK=1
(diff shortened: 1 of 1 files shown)


tillmo:
good. But how will the choice between `WITH_MAKE_CHECK=1` and `WITH_HETS_LIB=1` happen?

eugenk:
This is done by the build matrix:
```yaml
matrix:
  include:
    - env: WITH_MAKE_CHECK=1
    - env: WITH_HETS_LIB=1
```
This is actually intended for switching versions as in:
```yaml
matrix:
  include:
    - rvm: 2.3.3
      postgresql: 9.4.0
    - rvm: 2.3.3
      postgresql: 9.6.0
    - rvm: 2.4.1
      postgresql: 9.4.0
    - rvm: 2.4.1
      postgresql: 9.6.0
```
but it's possible to set environment variables as well. Each array item in `matrix.include` represents a job in the Travis build.

See the travis builds of this pull request for details.

eugenk:
This pull request is a dirty hack around the Travis-CI system and definitely not intended by the creators. Maybe this will be even prevented in the future. For now, this works, though.

I'm not sure if the caches will work properly. Both builds (the one of the push and the one of the pull request) failed to download the cache.

PR #1744, opened 2017-09-10
better error message for Logic.Logic.inclusion

tillmo:
the error message informs about the context that an inclusion is being constructed, and also prints source and target signature

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,10 @@ before_script:
       git clone --depth=1 https://github.com/spechub/Hets-lib.git $HOME/Hets-lib
     fi
 script:
+  # For some reason, Travis CI sets this environment variable, which breaks the
+  # tests (_JAVA_OPTIONS=-Xmx2048m -Xms512m)
+  - unset _JAVA_OPTIONS
+
   # Prepare to run Hets
   - export HETS_MAGIC=$PWD/magic/hets.magic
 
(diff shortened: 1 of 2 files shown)


tillmo:
@eugenk, do you understand the error message at https://travis-ci.org/spechub/Hets/jobs/273800782 ?

tillmo:
hm, now both checks fail...

mcodescu:
This is strange, the modifications look completely fine to me.

eugenk:
Travis CI sets this environment variable, which breaks the test suite. Both checks failed because I made a mistake in the `.travis.yml`. This is fixed my last commit (by `--amend`).

I have no idea why this the error came with this pull request because I can see no change that might have introduced this problem. Perhaps Travis changed things on their end.

jelmd:
Such a change would make sense, probably because lot of test envs do not set it by its own and thus wasting huge amounts of RAM/causing memory pressure without actually needing that much. IIRC, per default the JVM grows "its mem pool" 'til max., before it tries a little bit harder to garbage collect and re-use portions of the mem-pool. And IIRC., at least for Sun/Oracle JDKs the max. mem default is 4GB (I guess, for OpenJDK it is the same or more) ...

tillmo:
OK, now everything is fine. @mcodescu, can you please have a look?

mcodescu:
:+1:

Issue #1655, opened 2016-08-02
Problem with combination in OWL


tillmo:
With
```
logic OWL 
ontology Features = 
 Class:    Line 
end
ontology Shapes = 
 Class:    Line  
end
alignment RulesBase : Features to Shapes  = 
    Line =  Line
end
ontology Rules = 
  combine RulesBase
then
  Class: ShortLine SubClassOf: Line
end 
```
I get
```
/home/till/temp/verif2.dol:14.32-14.35: *** Error:
unknown Class: 'Line'
hets: user error (Stopped due to errors)
```

mcodescu:
The code on the main branch of Hets must be updated: currently the theory of the combination is:

```
Class: Line_source
Class: Line_target
       EquivalentTo: Line_source
```

mcodescu:
The main branch is up to date, and the issue is similar to #1597. I have implemented as a fix for the issue there a check whether a name of a left symbol appears as name of some right symbol, and if that's not the case, we can keep its original name. The problem with the example above is that we no longer use signature morphisms to encode equalities, but we have equivalences instead. Then we can't have `Line` equivalent to `Line`, but we must have two symbols with generated names. Thus, the solution for #1597 or an improvement of it will not work here. What we need is a decision whether we keep the generated names with `_source` and `_target` or we want to have other names.

tillmo:
once this has been resolved, add DOL/align2.dol and DOL/Alignment.dol to [test/hets-lib-testfiles](https://github.com/spechub/Hets/blob/0c1c431111e768803ecc7ac9c2725bec9b6787b4/test/hets-lib-testfiles)

Issue #1597, opened 2016-03-14
improve names in bridge ontology for alignments


tillmo:
Previously,
```
logic OWL

ontology Woman = Class: Woman 
ontology Person = Class: Person
alignment Alignment1 : Woman to Person =
  Woman < Person
end

ontology AlignedOntology1 =
  combine Alignment1
end
```
in the colimit lead to
```
Class: Person
Class: Woman
       SubClassOf: Person
```
Now we get
```
Class: Person_target
Class: Woman_source
       SubClassOf: Person_target
```
I recall that you have told me why this is so, but still I want to discuss if there is a way to get the old version, which looks nicer.

tillmo:
once this has been resolved, add DOL/align2.dol and DOL/Alignment.dol to [test/hets-lib-testfiles](https://github.com/spechub/Hets/blob/0c1c431111e768803ecc7ac9c2725bec9b6787b4/test/hets-lib-testfiles)

mcodescu:
Another thing to have in mind is that the names of the symbol can be IRIs. Thus, we have to decide: 
 - what to compare: most likely the iriPath
 - if the IRIs don't match (or if only their iriPath matches), what components has the IRI in the bridge ontology
 - what to do with prefix maps: unite the prefix maps of the ontologies being aligned and rename clashes?

PR #1752, opened 2017-09-18
[WIP] 1691 dg calculus on database

eugenk:
This adds database write-support to Hets to be used by Ontohub. Read support is not yet intended.

----
# Important Warning
As soon as this is merged, Hets cannot be compiled without Stack.

----

# Review Guide

# Compile
* `git submodule update --init --recursive` to pull the submodules (own fork of persistent)
* `make restack` to compile the new dependencies
* `make` to compile Hets

# Setup
* Install PostgreSQL, MySQL and SQLite
* Drop the database if it exists and create a new one:
  * MySQL (this is deactivated by not importing the corresponding Haskell module): `mysql -u root -e 'drop database if exists hets_development;'; mysql -u root -e 'create database hets_development;';`
  * PostgreSQL: `psql -U postgres -c 'drop database hets_development;'; psql -U postgres -c 'create database hets_development;';`
  * SQLite: just delete the database file.
* Run the database migrations and store the logic graph. This needs to be done only once for a database (or whenever the logic graph is expanded by new logics/sublogics):
    ```
    # MySQL (this is deactivated by not importing the corresponding Haskell module):
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_mysql.yml --database-subconfig=development --logic-graph
    # PostgreSQL:
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_postgresql.yml --database-subconfig=development --logic-graph
    # SQLite:
    ./hets --verbose=2 --output-types=db --database-file=/tmp/myHetsDatabase.db --logic-graph
    ```

# Run
* Hets-standalone:
    ```
    # MySQL (this is deactivated by not importing the corresponding Haskell module):
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_mysql.yml --database-subconfig=development --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    # PostgreSQL:
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_postgresql.yml --database-subconfig=development --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    # SQLite:
    ./hets --verbose=2 --output-types=db --database-file=/tmp/myHetsDatabase.db --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    ```
* Called by Ontohub:
    ```
    # MySQL (this is deactivated by not importing the corresponding Haskell module):
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_mysql.yml --database-subconfig=development --database-fileversion-id=5 --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    # PostgreSQL:
    ./hets --verbose=2 --output-types=db --database-config=Presistence/config/database_postgresql.yml --database-subconfig=development --database-fileversion-id=5 --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    # SQLite:
    ./hets --verbose=2 --output-types=db --database-file=/tmp/myHetsDatabase.db --database-fileversion-id=5 --hets-libdirs=$HETS_LIB $HETS_LIB/Basic/RelationsAndOrders.casl
    ```

# Arguments
* `--verbose=2` configures how many `Diagnose`s are saved in the database. These are hints, warnings errors and debug messages.
* `--output-types=db` tells Hets to write into the database
* `--database-config=Persistence/database_postgresql.yml` specifies the path to a Rails-compatible database configuration file. There are example files for PostgreSQL and MySQL in the Persistence directory.
* `--database-subconfig=development` tells Hets which key to use in the config file. Supported values: `production`, `development`, `test`
* `--hets-libdirs=$HETS_LIB` is used for creating LocId values. It will always be set to the repository root if Hets is called by Ontohub.
* `--database-fileversion-id=5` tells Hets which FileVersion to associate the LocIds with. This parameter is only used by Ontohub and is not supposed to be passed if Hets is used as a standalone tool. Please do not analyse different files with the same `--database-fileversion-id` parameter. Each file needs to be associated with a different FileVersion. Editing a file also makes it different.
* `--database-reanalyze` causes a file (FileVersion) to be reanalysed. This parameter is only used by Ontohub and is not supposed to be passed if Hets is used as a standalone tool. If the parameter `--database-fileversion-id` is not passed, Hets assumes that `--database-reanalyze` is set. Otherwise, it is not set by default.
* `$HETS_LIB/Basic/RelationsAndOrders.casl` is the file to analyse

# Code Review
There is a plethora of new code in this branch, but much of it does not need an in-depth review.
* `Persistence/Schema.hs` and `Persistence/Schema/*.hs` contain several hundred LOC that only describe the database schema. It is tailored to the already reviewed schema of Ontohub and can be skimmed.
* `Persistence/db_mapping.txt` can be skipped in favour of `Persistence/Readme.md`. These two contain basically the same information, but the latter is more reader-friendly.

Other things about the code:
* `Static/ToJson.hs`: The changes only export some of the existing functions
* `test/*` are only used by Travis - They set up and tear down the database and run the above-mentioned commands.
* `Persistence/DevGraph/Cleaning.hs` exports only one function to delete the database entries that need to be removed for a call with `--database-reanalyze`.

The most important business logic is located in
* `Persistence/LogicGraph.hs`: Migrating the logic graph
* `Persistence/DevGraph.hs`: Saving/Replacing the development graph
* `Persistence/Diagnosis.hs`: Saving Debug/Hint/Warn/Error messages
* `Persistence/FileVersion.hs`: Finding or creating a FileVersion

### DIFF
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -48,6 +48,7 @@ import qualified Common.Lib.Rel as Rel
 import Common.LibName
 import Common.Result (maybeResult)
 import Common.ResultT (runResultT)
+import Common.Utils
 import Driver.Options
 import qualified Common.OrderedMap as OMap
 import Logic.Comorphism
@@ -598,28 +599,45 @@ findOrCreateSymbolMappings libEnv dbCache doSave libName edge
   signatureMorphismKey gMorphism = case gMorphism of
     GMorphism { gMorphismComor = cid
               , gMorphismMor = morphism
+              , gMorphismSign = extSign@ExtSign { plainSign = sign }
               } ->
-      let lid = targetLogic cid in do
-        symbolMappingKeys <-
-          mapM (findOrCreateSymbolMapping libEnv dbCache doSave libName edge
-                  signatureMorphismKey lid
-               ) $ Map.toList $ symmap_of lid morphism
-        return (symbolMappingKeys, dbCache)
+      let sourceLid = sourceLogic cid
+          targetLid = targetLogic cid
+          symbolMap = symmap_of targetLid morphism
+      in  do
+            symbolMappingKeys <-
+              concatMapM
+                (\ sourceSymbol ->
+                  concatMapM
+                    (\ translatedSymbol ->
+                      case Map.lookup translatedSymbol symbolMap of
+                        Nothing -> return []
+                        Just targetSymbol -> do
+                            symbolMappingKey <-
+                              findOrCreateSymbolMapping libEnv dbCache doSave
+                                libName edge signatureMorphismKey sourceLid
+                                targetLid sourceSymbol targetSymbol
+                            return [symbolMappingKey]
+                    ) $ Set.toList $ map_symbol cid sign sourceSymbol
+                ) $ Set.toList $ symset_of sourceLid $ plainSign extSign
+            return (symbolMappingKeys, dbCache)
 
 findOrCreateSymbolMapping :: ( MonadIO m
-                             , Sentences lid sentence sign morphism symbol
+                             , Sentences lid1 sentence1 sign1 morphism1 symbol1
+                             , Sentences lid2 sentence2 sign2 morphism2 symbol2
                              )
                           => LibEnv -> DBCache -> Bool -> LibName -> (Int, Int)
-                          -> SignatureMorphismId -> lid -> (symbol, symbol)
+                          -> SignatureMorphismId
+                          -> lid1 -> lid2 -> symbol1 -> symbol2
                           -> DBMonad m SymbolMappingId
 findOrCreateSymbolMapping libEnv dbCache doSave libName (sourceId, targetId)
-  signatureMorphismKey lid (sourceSymbol, targetSymbol) = do
+  signatureMorphismKey sourceLid targetLid sourceSymbol targetSymbol = do
   let sourceKey =
-        fromJust $ findSymbolInCache libEnv libName sourceId lid sourceSymbol
+        fromJust $ findSymbolInCache libEnv libName sourceId sourceLid sourceSymbol
           dbCache
 
   let targetKey =
-        fromJust $ findSymbolInCache libEnv libName targetId lid targetSymbol
+        fromJust $ findSymbolInCache libEnv libName targetId targetLid targetSymbol
           dbCache
 
   symbolMappingM <-
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -16,6 +16,7 @@ Portability :  portable
 module Persistence.DevGraph (exportLibEnv) where
 
 import Persistence.Database
+import Persistence.Diagnosis
 import Persistence.FileVersion
 import Persistence.LogicGraph
 import Persistence.Schema as SchemaClass
@@ -45,7 +46,7 @@ import qualified Common.Lib.Graph as Graph (nodeLabel)
 import Common.Lib.Rel (Rel)
 import qualified Common.Lib.Rel as Rel
 import Common.LibName
-import Common.Result (maybeResult)
+import Common.Result (maybeResult, Diagnosis(..), DiagKind(..))
 import Common.ResultT (runResultT)
 import Common.Utils
 import Driver.Options
@@ -59,6 +60,7 @@ import Static.DgUtils
 import Static.GTheory
 import qualified Static.ToJson as ToJson
 
+import Control.Exception
 import Control.Monad (foldM, foldM_, when)
 import Control.Monad.IO.Class (MonadIO (..))
 import Data.Char (toLower)
@@ -96,17 +98,31 @@ emptyDBCache = DBCache { documentMap = Map.empty
 
 exportLibEnv :: HetcatsOpts -> LibEnv -> IO ()
 exportLibEnv opts libEnv = do
-  onDatabase (databaseConfig opts) $
-    setFileVersionState (databaseContext opts) EvaluationStateType.Processing
-  onDatabase (databaseConfig opts) $ do
-    let dependencyLibNameRel = getLibDepRel libEnv
-    let dependencyOrderedLibsSetL = Rel.depSort dependencyLibNameRel
-    dbCache1 <-
-      createDocuments opts libEnv emptyDBCache dependencyOrderedLibsSetL
-    createDocumentLinks dbCache1 dependencyLibNameRel
-    setFileVersionState (databaseContext opts)
-      EvaluationStateType.FinishedSuccessfully
-    return ()
+  onDatabase dbConfig $
+    setFileVersionState dbContext EvaluationStateType.Processing
+  Control.Exception.catch export saveErrorAndRethrow
+  where
+    dbConfig = databaseConfig opts
+    dbContext = databaseContext opts
+
+    export :: IO ()
+    export = onDatabase dbConfig $ do
+      let dependencyLibNameRel = getLibDepRel libEnv
+      let dependencyOrderedLibsSetL = Rel.depSort dependencyLibNameRel
+      dbCache1 <-
+        createDocuments opts libEnv emptyDBCache dependencyOrderedLibsSetL
+      createDocumentLinks dbCache1 dependencyLibNameRel
+      setFileVersionState dbContext EvaluationStateType.FinishedSuccessfully
+      return ()
+
+    saveErrorAndRethrow :: SomeException -> IO ()
+    saveErrorAndRethrow exception = do
+      let message = show exception
+      saveDiagnoses dbConfig dbContext 5 [Diag { diagKind = Error
+                                               , diagString = message
+                                               , diagPos = nullRange
+                                               }]
+      fail message
 
 createDocuments :: MonadIO m
                 => HetcatsOpts -> LibEnv -> DBCache -> [Set LibName]
@@ -633,13 +649,13 @@ findOrCreateSymbolMapping :: ( MonadIO m
                           -> DBMonad m SymbolMappingId
 findOrCreateSymbolMapping libEnv dbCache doSave libName (sourceId, targetId)
   signatureMorphismKey sourceLid targetLid sourceSymbol targetSymbol = do
-  let sourceKey =
-        fromJust $ findSymbolInCache libEnv libName sourceId sourceLid sourceSymbol
-          dbCache
+  sourceKey <-
+    fromJustFail (symbolErrorMessage libEnv libName sourceId sourceSymbol sourceLid "source") $
+      findSymbolInCache libEnv libName sourceId sourceLid sourceSymbol dbCache
 
-  let targetKey =
-        fromJust $ findSymbolInCache libEnv libName targetId targetLid targetSymbol
-          dbCache
+  targetKey <-
+    fromJustFail (symbolErrorMessage libEnv libName targetId targetSymbol targetLid "target") $
+      findSymbolInCache libEnv libName targetId targetLid targetSymbol dbCache
 
   symbolMappingM <-
     selectFirst [ SymbolMappingSignatureMorphismId ==. signatureMorphismKey
@@ -656,6 +672,23 @@ findOrCreateSymbolMapping libEnv dbCache doSave libName (sourceId, targetId)
             , symbolMappingTargetId = targetKey
             }
       else fail "Persistence.DevGraph.findOrCreateSymbolMapping: not found"
+
+symbolErrorMessage :: Sentences lid sentence sign morphism symbol
+                   => LibEnv -> LibName -> Node -> symbol -> lid -> String
+                   -> String
+symbolErrorMessage libEnv libName nodeId symbol lid end =
+  let (libNameDereferenced, nodeIdDereferenced) = dereferenceNodeId libEnv libName nodeId
+      fullName = show $ fullSymName lid symbol
+      symbolKind = symKind lid symbol
+      nodeName = show $ getName $ dgn_name $ getNodeLabel libEnv libName nodeId
+  in  "Could not find " ++ end ++ " symbol for SymbolMapping:"
+      ++ "\nsymbol name: " ++ fullName
+      ++ "\nsymbol kind: " ++ symbolKind
+      ++ "\nnode name: " ++ nodeName
+      ++ "\nnode ID: " ++ show nodeIdDereferenced
+      ++ "\nlibrary name: " ++ show libNameDereferenced
+      ++ "\nThis is a system error. Please report it at https://github.com/spechub/Hets/issues"
+
 
 findOrCreateSerializationM :: MonadIO m
                            => LanguageId -> Maybe IRI
@@ -1227,3 +1260,9 @@ createDocumentLinks dbCache dependencyLibNameRel =
                           , documentLinkTargetId = targetKey
                           }
       return ()
+
+
+fromJustFail :: MonadIO m => String -> Maybe a -> m a
+fromJustFail message maybeValue = case maybeValue of
+  Just value -> return value
+  Nothing -> fail message
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -286,8 +286,7 @@ createAllOmsOfDocument :: MonadIO m
                        -> Bool -> DGraph -> GlobalAnnos -> LibName
                        -> Entity LocIdBase -> DBMonad m DBCache
 createAllOmsOfDocument opts libEnv fileVersionKey dbCache0 doSave dGraph
-  globalAnnotations libName
-  documentLocIdBase@(Entity _ documentLocIdBaseValue) =
+  globalAnnotations libName documentLocIdBase =
   let labeledNodes = labNodes $ dgBody dGraph
       labeledEdges = labEdges $ dgBody dGraph
   in  do
@@ -299,8 +298,8 @@ createAllOmsOfDocument opts libEnv fileVersionKey dbCache0 doSave dGraph
           ) dbCache0 labeledNodes
         foldM
           (\ dbCacheAcc labeledEdge ->
-            createMapping opts libEnv fileVersionKey dbCacheAcc
-              globalAnnotations libName documentLocIdBaseValue labeledEdge
+            createMapping opts libEnv fileVersionKey dbCacheAcc doSave
+              globalAnnotations libName documentLocIdBase labeledEdge
           ) dbCache1 labeledEdges
 
 findOrCreateOMSM :: MonadIO m
@@ -1035,11 +1034,12 @@ associateSymbolsOfSignature libEnv dbCache libName nodeId lid extSign signatureK
        return dbCache
 
 createMapping :: MonadIO m
-              => HetcatsOpts -> LibEnv -> FileVersionId -> DBCache
-              -> GlobalAnnos -> LibName -> LocIdBase -> (Int, Int, DGLinkLab)
-              -> DBMonad m DBCache
-createMapping opts libEnv fileVersionKey dbCache globalAnnotations
-  libName documentLocIdBaseValue (sourceId, targetId, linkLabel) = do
+              => HetcatsOpts -> LibEnv -> FileVersionId -> DBCache -> Bool
+              -> GlobalAnnos -> LibName -> Entity LocIdBase
+              -> (Int, Int, DGLinkLab) -> DBMonad m DBCache
+createMapping opts libEnv fileVersionKey dbCache doSave globalAnnotations
+  libName documentLocIdBase@(Entity _ documentLocIdBaseValue)
+  (sourceId, targetId, linkLabel) = do
   (sourceOMSKey, sourceSignatureKey) <-
     case Map.lookup (libName, sourceId) $ nodeMap dbCache of
       Just (omsKey, signatureKey) -> return (omsKey, signatureKey)
@@ -1071,13 +1071,14 @@ createMapping opts libEnv fileVersionKey dbCache globalAnnotations
         fmap Just $ createConservativityStatus consStatus
       _ -> return Nothing
 
-    freenessParameterOMSKeyM <- case dgl_type linkLabel of
-      FreeOrCofreeDefLink _ (JustNode NodeSig { getNode = nodeId }) ->
-        case Map.lookup (libName, nodeId) $ nodeMap dbCache of
-          Just (freenessParameterOMSKey, _) ->
-            return $ Just freenessParameterOMSKey
-          Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find freeness parameter node with ID " ++ show (libName, nodeId))
-      _ -> return Nothing
+    (freenessParameterOMSKeyM, dbCache2) <- case dgl_type linkLabel of
+      FreeOrCofreeDefLink _ (JustNode NodeSig { getNode = nodeId }) -> do
+        let nodeLabel = getNodeLabel libEnv libName nodeId
+        (freenessParameterOMSKey, _, dbCache') <-
+          findOrCreateOMS opts libEnv fileVersionKey dbCache1 doSave
+            globalAnnotations libName documentLocIdBase (nodeId, nodeLabel)
+        return (Just freenessParameterOMSKey, dbCache')
+      _ -> return (Nothing, dbCache)
 
     freenessParameterLanguageKeyM <- case dgl_type linkLabel of
       FreeOrCofreeDefLink _ (EmptyNode (Logic.Logic lid)) ->
@@ -1103,7 +1104,7 @@ createMapping opts libEnv fileVersionKey dbCache globalAnnotations
           , mappingPending = dglPending linkLabel
           }
     insertEntityMany [Entity (toSqlKey $ fromSqlKey mappingKey) mapping]
-    return dbCache1
+    return dbCache2
   where
     mappingOriginOfLinkLabel :: MappingOrigin
     mappingOriginOfLinkLabel = case dgl_origin linkLabel of
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -1043,12 +1043,12 @@ createMapping opts libEnv fileVersionKey dbCache globalAnnotations
   (sourceOMSKey, sourceSignatureKey) <-
     case Map.lookup (libName, sourceId) $ nodeMap dbCache of
       Just (omsKey, signatureKey) -> return (omsKey, signatureKey)
-      Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find source node with ID " ++ show sourceId)
+      Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find source node with ID " ++ show (libName, sourceId))
 
   (targetOMSKey, targetSignatureKey) <-
     case Map.lookup (libName, targetId) $ nodeMap dbCache of
       Just (omsKey, signatureKey) -> return (omsKey, signatureKey)
-      Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find target node with ID " ++ show targetId)
+      Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find target node with ID " ++ show (libName, targetId))
 
   (signatureMorphismKey, _, dbCache1) <-
     findOrCreateSignatureMorphism opts libEnv dbCache True libName
@@ -1076,7 +1076,7 @@ createMapping opts libEnv fileVersionKey dbCache globalAnnotations
         case Map.lookup (libName, nodeId) $ nodeMap dbCache of
           Just (freenessParameterOMSKey, _) ->
             return $ Just freenessParameterOMSKey
-          Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find freeness parameter node with ID " ++ show nodeId)
+          Nothing -> fail ("Persistence.DevGraph.createMapping: Could not find freeness parameter node with ID " ++ show (libName, nodeId))
       _ -> return Nothing
 
     freenessParameterLanguageKeyM <- case dgl_type linkLabel of
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -115,8 +115,8 @@ createDocuments :: MonadIO m
 createDocuments opts libEnv dbCache0 dependencyOrderedLibsSetL = do
   fileVersion <- getFileVersion $ databaseContext opts
   dbCache1 <-
-    createDocumentsInDependencyRelation opts libEnv fileVersion dbCache0
-      dependencyOrderedLibsSetL
+    createDocumentsInDependencyRelation opts libEnv fileVersion dbCache0 $
+      reverse dependencyOrderedLibsSetL
   createDocumentsThatAreIndependent opts libEnv fileVersion dbCache1
 
 createDocumentsInDependencyRelation :: MonadIO m
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -800,7 +800,7 @@ findOrCreateSymbol :: ( MonadIO m
 findOrCreateSymbol libEnv fileVersionKey dbCache doSave libName nodeId
   (Entity omsKey omsLocIdBaseValue) lid symbol =
     let name = show $ sym_name lid symbol
-        fullName = show $ fullSymName lid symbol
+        fullName = fullSymName lid symbol
         symbolKind = symKind lid symbol
         locId = locIdBaseLocId omsLocIdBaseValue ++ "//" ++ symbolKind ++ "/" ++ name
     in  if symbolIsCached libEnv libName nodeId lid symbol dbCache
--- a/Persistence/DevGraph.hs
+++ b/Persistence/DevGraph.hs
@@ -179,13 +179,13 @@ createDocument opts libEnv parentFileVersion dbCache0 libName =
                     , locIdBaseLocId = locId
                     }
               documentKey <- insert documentLocIdBaseValue
-              insert Document
-                { documentLocIdBaseId = documentKey
-                , documentDisplayName = displayName
-                , documentName = name
-                , documentLocation = location
-                , documentVersion = version
-                }
+              let document = Document
+                    { documentDisplayName = displayName
+                    , documentName = name
+                    , documentLocation = location
+                    , documentVersion = version
+                    }
+              insertEntityMany [Entity (toSqlKey $ fromSqlKey documentKey) document]
               return (True, documentKey, documentLocIdBaseValue)
           let dbCache1 = addDocumentToCache libName documentKey dbCache0
           createAllOmsOfDocument opts libEnv fileVersionKey dbCache1 doSave dGraph
@@ -415,27 +415,27 @@ createOMS opts libEnv fileVersionKey dbCache0 doSave globalAnnotations libName
                   , locIdBaseLocId = locId
                   }
             omsKey <- insert omsLocIdBaseValue
-            insert OMS
-              { oMSLocIdBaseId = omsKey
-              , oMSDocumentId = documentKey
-              , oMSLanguageId = languageKey
-              , oMSLogicId = logicKey
-              , oMSSerializationId = serializationKey
-              , oMSSignatureId = signatureKey
-              , oMSNormalFormId = normalFormKeyM
-              , oMSNormalFormSignatureMorphismId = normalFormSignatureMorphismKeyM
-              , oMSFreeNormalFormId = freeNormalFormKeyM
-              , oMSFreeNormalFormSignatureMorphismId = freeNormalFormSignatureMorphismKeyM
-              , oMSOrigin = omsOrigin
-              , oMSConsStatusId = consStatusKey
-              , oMSNameFileRangeId = nodeNameRangeKey
-              , oMSDisplayName = displayName
-              , oMSName = name
-              , oMSNameExtension = nameExtension
-              , oMSNameExtensionIndex = nameExtensionIndex
-              , oMSLabelHasHiding = labelHasHiding nodeLabel
-              , oMSLabelHasFree = labelHasFree nodeLabel
-              }
+            let oms = OMS
+                  { oMSDocumentId = documentKey
+                  , oMSLanguageId = languageKey
+                  , oMSLogicId = logicKey
+                  , oMSSerializationId = serializationKey
+                  , oMSSignatureId = signatureKey
+                  , oMSNormalFormId = normalFormKeyM
+                  , oMSNormalFormSignatureMorphismId = normalFormSignatureMorphismKeyM
+                  , oMSFreeNormalFormId = freeNormalFormKeyM
+                  , oMSFreeNormalFormSignatureMorphismId = freeNormalFormSignatureMorphismKeyM
+                  , oMSOrigin = omsOrigin
+                  , oMSConsStatusId = consStatusKey
+                  , oMSNameFileRangeId = nodeNameRangeKey
+                  , oMSDisplayName = displayName
+                  , oMSName = name
+                  , oMSNameExtension = nameExtension
+                  , oMSNameExtensionIndex = nameExtensionIndex
+                  , oMSLabelHasHiding = labelHasHiding nodeLabel
+                  , oMSLabelHasFree = labelHasFree nodeLabel
+                  }
+            insertEntityMany [Entity (toSqlKey $ fromSqlKey omsKey) oms]
             return $ Entity omsKey omsLocIdBaseValue
 
         dbCache6 <- case gTheory of
@@ -782,14 +782,14 @@ findOrCreateSymbol libEnv fileVersionKey dbCache doSave libName nodeId
                 , locIdBaseKind = Enums.Symbol
                 , locIdBaseLocId = locId
                 }
-              insert Symbol
-                { symbolLocIdBaseId = symbolKey
-                , symbolOmsId = omsKey
-                , symbolFileRangeId = Nothing
-                , symbolSymbolKind = symbolKind
-                , symbolName = name
-                , symbolFullName = Text.pack fullName
-                }
+              let symbolValue = Symbol
+                    { symbolOmsId = omsKey
+                    , symbolFileRangeId = Nothing
+                    , symbolSymbolKind = symbolKind
+                    , symbolName = name
+                    , symbolFullName = Text.pack fullName
+                    }
+              insertEntityMany [Entity (toSqlKey $ fromSqlKey symbolKey) symbolValue]
               return symbolKey
             else do
               symbolM <- selectFirst [ LocIdBaseLocId ==. locId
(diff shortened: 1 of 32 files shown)


eugenk:
There will be some more changes in the dependencies (persistent, persistent-*), but the Haskell code should be finalised now.

eugenk:
I have pulled my fork of persistent as a git submodule into this repository. It is necessary because otherwise, we cannot use PostgreSQL enumeration types. I also added a review guide to the first post.

This is *not* ready for review. As soon as this is ready, I will post it here and remove the [WIP] tag from the title.

tillmo:
Actually, I still get some error:
```
[748 of 817] Compiling Persistence.Schema.ReasoningStatusOnReasoningAttemptType ( Persistence/Schema/ReasoningStatusOnReasoningAttemptType.hs, Persistence/Schema/ReasoningStatusOnReasoningAttemptType.o )

Persistence/Schema/ReasoningStatusOnReasoningAttemptType.hs:3:1: error:
    Failed to load interface for ‘Database.Persist.TH’
```

eugenk:
I was wrong with `git submodule update`. Instead, it is `git submodule update --init --recursive`.

eugenk:
This should be ready for review now.

Persistence/DevGraph.hs
```diff
@@ -0,0 +1,1216 @@
...
+        ++ "WHERE (file_versions.path = ? "
+        ++ "       AND backreference.id = ?);")
+
+  in  if isInLibdir
+      then do
+            results <-
+              rawSql queryString [ toPersistValue path
+                                 , toPersistValue fileVersionKey
+                                 ]
+            return $ head results
```
tillmo: With tracing, I could locate the following error to be caused by `results` being the empty list:
```
till@mark:~/projects/ontohub/ontohub-backend$ hets --verbose=2 --output-types=db --database-config=config/database.yml --database-subconfig=development --hets-libdirs=$HETS_LIB --database-fileversion-id=5 $HETS_LIB/Basic/RelationsAndOrders.casl
### Warning: set HETS_MAGIC to a proper magic file
Reading file /home/till/CASL/Hets-lib/Basic/RelationsAndOrders.casl
Analyzing library Basic/RelationsAndOrders version 1.0
Downloading Basic/Numbers ...
### Warning: set HETS_MAGIC to a proper magic file
Reading file /home/till/CASL/Hets-lib/Basic/Numbers.casl
Analyzing library Basic/Numbers version 1.0
Analyzing spec Basic/Numbers?Nat
Analyzing spec Basic/Numbers?Int
Analyzing spec Basic/Numbers?Rat
Analyzing spec Basic/Numbers?DecimalFraction
... loaded Basic/Numbers
Analyzing spec Basic/RelationsAndOrders?Relation
Analyzing spec Basic/RelationsAndOrders?ReflexiveRelation
Analyzing spec Basic/RelationsAndOrders?IrreflexiveRelation
Analyzing spec Basic/RelationsAndOrders?SymmetricRelation
Analyzing spec Basic/RelationsAndOrders?AsymmetricRelation
Analyzing spec Basic/RelationsAndOrders?AntisymmetricRelation
Analyzing spec Basic/RelationsAndOrders?TransitiveRelation
Analyzing spec Basic/RelationsAndOrders?SimilarityRelation
Analyzing spec Basic/RelationsAndOrders?PartialEquivalenceRelation
Analyzing spec Basic/RelationsAndOrders?EquivalenceRelation
Analyzing spec Basic/RelationsAndOrders?PreOrder
Analyzing spec Basic/RelationsAndOrders?StrictOrder
Analyzing spec Basic/RelationsAndOrders?PartialOrder
Analyzing spec Basic/RelationsAndOrders?TotalOrder
Analyzing spec Basic/RelationsAndOrders?StrictTotalOrder
Analyzing spec Basic/RelationsAndOrders?RightUniqueRelation
Analyzing spec Basic/RelationsAndOrders?LeftTotalRelation
Analyzing spec Basic/RelationsAndOrders?BooleanAlgebra
Analyzing spec Basic/RelationsAndOrders?ExtPartialOrder
Analyzing spec Basic/RelationsAndOrders?ExtTotalOrder
Analyzing spec Basic/RelationsAndOrders?ExtBooleanAlgebra
Analyzing spec Basic/RelationsAndOrders?RichPartialOrder
Analyzing spec Basic/RelationsAndOrders?RichTotalOrder
Analyzing spec Basic/RelationsAndOrders?RichBooleanAlgebra
Analyzing view Basic/RelationsAndOrders?TotalOrder_in_Nat
Analyzing view Basic/RelationsAndOrders?TotalOrder_in_Int
Analyzing view Basic/RelationsAndOrders?TotalOrder_in_Rat
Analyzing view Basic/RelationsAndOrders?PartialOrder_in_ExtBooleanAlgebra
hets: Prelude.head: empty list
```

mcodescu:
With this last commit, I now get
```
...
Analyzing view Basic/RelationsAndOrders?PartialOrder_in_ExtBooleanAlgebra
hets: Language not found in the database: CASL
```
I have run recreate with seed:
```
rails db:recreate:seed
Invoker does not seem to be running
Invoker does not seem to be running
[sequel] Dropped database 'ontohub_development'
[sequel] Created database 'ontohub_development'
Invoker does not seem to be running
Invoker does not seem to be running
```

eugenk:
Sorry, I split exporting the logic graph from the analysis into separate calls. Please run 
```text
hets-self --verbose=2 --output-types=db --database-config=${ontohub-backend}/config/database.yml --database-subconfig=development --logic-graph
```
once. I added this to the review guide.

eugenk:
There is one more change to come: The behaviour about finding already saved documents.

mcodescu:
Great, now I can run hets successfully on this branch.

tillmo:
With the following heterogeneous OWL2 and CASL example I got:
```
hets --verbose=2 --output-types=db --database-config=config/database.yml --database-subconfig=development --logic-graph
till@mark:~/projects/ontohub/ontohub-backend$ hets --verbose=2 --output-types=db --database-config=config/database.yml --database-subconfig=development --hets-libdirs=$HETS_LIB --database-fileversion-id=5 /home/till/CASL/Hets-lib/Ontology/Examples/Cats.het 
### Warning: set HETS_MAGIC to a proper magic file
Reading file /home/till/CASL/Hets-lib/Ontology/Examples/Cats.het
Analyzing file /home/till/CASL/Hets-lib/Ontology/Examples/Cats.het as library Cats
logic OWL
Analyzing spec Cats?Predators
logic CASL
Analyzing spec Cats?Prey
logic CASL
Analyzing spec Cats?Animals
hets: SqlError {sqlState = "23505", sqlExecStatus = FatalError, sqlErrorMsg = "duplicate key value violates unique constraint \"loc_id_bases_file_version_id_kind_loc_id_index\"", sqlErrorDetail = "Key (file_version_id, kind, loc_id)=(5, Symbol, Ontology/Examples/Cats.het//Prey__E1//Thing) already exists.", sqlErrorHint = ""}
```

tillmo:
With `Basic/LinearAlgebra_II.casl`, I get
```
[...]
Analyzing spec Basic/LinearAlgebra_II?RichFreeAlgebra
Analyzing view Basic/LinearAlgebra_II?FreeVectorSpace_in_VSWithBase
Analyzing view Basic/LinearAlgebra_II?FreeAlgebra_in_Polynomial
hets: Persistence.DevGraph.createOMS: OMS not foundBasic/StructuredDatatypes.casl//Set__PU1
```

tillmo:
here a small example producing the sql error:
```
library TestBasic

spec Datatypes =
     type Data1 ::= a | b | c;

     types Tree ::= Leaf (Data1) | Forest;
           Forest ::= Nil | Cons (Tree;Forest)
  
end
```

tillmo:
Even shorter:
```
spec Datatypes =
     types Tree ::= Leaf | Forest;
           Forest ::= Nil | Cons (Tree;Forest)
end

Reading file /home/till/temp/d.casl
Analyzing file /home/till/temp/d.casl as library d
Analyzing spec d?Datatypes
hets: SqlError {sqlState = "23505", sqlExecStatus = FatalError, sqlErrorMsg = "duplicate key value violates unique constraint \"loc_id_bases_file_version_id_kind_loc_id_index\"", sqlErrorDetail = "Key (file_version_id, kind, loc_id)=(5, Symbol, /home/till/temp/d.casl//Datatypes//Forest) already exists.", sqlErrorHint = ""}
```

tillmo:
unit specs should be ignored. Currently, they lead to an error. For example,
```
spec sp1 =
  sort s
end
unit spec usp1 = sp1 * sp1 -> sp1
end
```
leads to
```
hets: Maybe.fromJust: Nothing
```

tillmo:
A heterogeneous specification like
```
logic Propositional
spec sp = 
prop p
. p %implied
end

logic CASL
spec sp1 = sp 
  then sort t
end
```
leads to an error:
```
hets: Maybe.fromJust: Nothing
```

tillmo:
The previous example is also a smaller example for the `OMS not found` error. First, create a file with specification `sp` only. Then, add `sp1` and analyse again. Result:
```
hets: Persistence.DevGraph.createOMS: OMS not found/home/till/temp/h.casl//sp1__X
```

tillmo:
After solving the above issues, the Hets-lib examples should be added as tests (actually, this is how I found those bugs).

eugenk:
Considering https://github.com/spechub/Hets/pull/1752#issuecomment-336633747, we need to qualify symbols differently in the Loc/Id. We have the symbol "Forest" once with kind "sort" and once with kind "op". I suggest to make the Loc/Id of a symbol have this kind before the symbol name.

tillmo:
Agreed, this should become part of the Loc/Id. Note that #1596 will provide a useful IRI anyway.

eugenk:
About https://github.com/spechub/Hets/pull/1752#issuecomment-336636315:
It looks like unit specs don't contain symbols.

* for every DGNode, I print "DGNode <nodeId>"
* for every symbol in a node, I print
  * the index of the symbol (that is about to be saved and cached) in the symbolKeyMap
  * a hint that it is being saved and cached
  * the symbolKeyMap from the cache after the symbol was added
* for every SymbolMapping, I print what symbol is searched and where

There are four nodes, of which only the first two actually contain symbols. Then, the mapping 

```text
DGNode 0
findSymbolInCache (u,0,("sort","\"s\""))
findOrCreateSymbol ("sort","\"s\"")
newSymbolCache: fromList [((u,0,("sort","\"s\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 53}})]
findSymbolInCache (u,0,("sort","\"s\""))
DGNode 1
findSymbolInCache (u,1,("sort","\"s\""))
findOrCreateSymbol ("sort","\"s\"")
newSymbolCache: fromList [((u,0,("sort","\"s\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 53}}),((u,1,("sort","\"s\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 55}})]
DGNode 2
DGNode 3
DGNode 4
-- Now the SymbolMappings are about to be created. The second element in the tuple is the nodeId where the symbol should be located.
findSymbolInCache (u,0,("sort","\"s\""))
findSymbolInCache (u,1,("sort","\"s\""))
findSymbolInCache (u,0,("sort","\"s\""))
findSymbolInCache (u,2,("sort","\"s\""))
```

The symbol is searched in Node 2, but there is no symbol at all in Node 2. How can it be?

eugenk:
https://github.com/spechub/Hets/pull/1752#issuecomment-336636505:

This is the same:
```
DGNode 0 p?sp
findSymbolInCache (p,0,("prop","\"p\""))
newSymbolCache: fromList [((p,0,("prop","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 86}})]
findSymbolInCache (p,0,("prop","\"p\""))

DGNode 1 p?sp1
findSymbolInCache (p,1,("pred","\"p\""))
newSymbolCache: fromList [((p,0,("prop","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 86}}),((p,1,("pred","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 89}})]
findSymbolInCache (p,1,("pred","\"p\""))

DGNode 2 p?sp1
findSymbolInCache (p,2,("sort","\"t\""))
newSymbolCache: fromList [((p,0,("prop","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 86}}),((p,1,("pred","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 89}}),((p,2,("sort","\"t\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 91}})]
findSymbolInCache (p,2,("pred","\"p\""))
newSymbolCache: fromList [((p,0,("prop","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 86}}),((p,1,("pred","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 89}}),((p,2,("pred","\"p\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 92}}),((p,2,("sort","\"t\"")),LocIdBaseKey {unLocIdBaseKey = SqlBackendKey {unSqlBackendKey = 91}})]
findSymbolInCache (p,2,("pred","\"p\""))
findSymbolInCache (p,2,("sort","\"t\""))

sourceId: 0
targetId: 1
findSymbolInCache (p,0,("pred","\"p\""))
fromJust: Nothing
```

The symbol `p` is searched in Node 0, but the symbol doesn't exist there. It only exists in Node 1.

eugenk:
Please do not analyse different files with the same --database-fileversion-id parameter. Each file needs to be associated with a different FileVersion. Editing a file also makes it different. Sorry, I forgot to put it into the guide.

tillmo:
Hm, consider
```
spec sp =
  sort s
  ops c,d:s
then
  . c=d
end
```
Then one of the nodes does not have (local) symbols either.

tillmo:
the problem with unit specs is a separate issue, see #1760

eugenk:
Now the heterogeneous example should pass

eugenk:
I think making this completely independent of Ontohub is more work than I thought. We would need to replicate more of the database (repository, user) in order to make the not-null-constraints pass.

eugenk:
The last commit requires https://github.com/ontohub/ontohub-models/pull/113 to be merged.

tillmo:
Concerning the database issue, I suggest:
- if no file version is specified, one can specify a repository and then the file version is created in that repository
- if in this case no repository is specified either, the first repository is used
- if in this case there is no repository at all, a new repository is created, for the user that is specified
- if in this case there is no user specified, the first user is used
- if in this case there is no user at all, a new user is created
The creation processes should lead to issuing a warning

tillmo:
Now that #1762 has been merged, please rebase to master. This should solve the problem with parameterised unit specifications.

eugenk:
rebased on master

eugenk:
Now you can use an SQLite database. Just use `--database-file=/tmp/hets.sqlite` instead of `--database-config=${ontohub-backend}/config/database.yml --database-subconfig=development` in the command line parameters.

eugenk:
The job on travis fails at Hets-lib/Calculi/Space/DRA_HO.het, trying to find the symbol `0` with kind `op` from the node number 5 in the library `HasCASL/Real version 0.2´.
```text
source: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Elem})}
target: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Elem})}
findSymbolInCache: (HasCASL/Real version 0.2,4,("op","\"0\""))
findSymbolInCache: (HasCASL/Real version 0.2,5,("op","\"0\""))
```

Also, the travis job takes too long. I assume because SQLite is slow.

eugenk:
Interestingly, there are two DGNodes with the name "HasCASL/Real?Field" (IDs 4 and 5) and the symbol "0" from ID 4 is mapped to the symbol "0" from ID 5. This fails because the latter doesn't exist.

tillmo:
I could minimise the problem from `Hets-lib/Calculi/Space/DRA_HO.het` as follows:
```
from test get Field |-> BasicField
spec Field = BasicField 
```
with `test.casl` (placed in `$HETS_LIB`) containing
```
spec Monoid = sort Elem
spec Field = Monoid 
```

eugenk:
With
```
from test get Field |-> BasicField
spec Sp = sort t
spec Field = Sp
```
Hets has not a single symbol inside the last Node (`Field`).

However, with
```
from test get Field |-> BasicField
spec Sp = sort t
spec Field2 = Sp
```
Hets has a symbol (`t`, a sort) inside the last Node (`Field2`). Is this a bug in the persistence layer or in the remainder of Hets?

tillmo:
Hm, in the DGXML (with `-o xml`) I cannot see any difference. Also, in the GUI, there is no difference: both nodes are locally empty, but have a sort t imported (from the Sp node).

eugenk:
Taking a closer look at the problem, I found that I had the dependency relation backwards. This is fixed by d479f7b877ed0d42a6380efc094279f8de560203

eugenk:
Now, travis only fails because of the time it takes to run the tests.

tillmo:
so all Hets-lib files that usually are tested also go through with the database?
I think it would be good to have all of them as tests, either by splitting the travis jobs even more, or by using a faster db like mysql (the latter is probably easier?)

eugenk:
> so all Hets-lib files that usually are tested also go through with the database?

Actually, not. I just ran this locally and got this:

```text
source: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
target: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
findSymbolInCache: (Basic/Numbers version 1.0,2,("op","\"0\""))
findSymbolInCache: (Invoice_corrected,47,("op","\"0\""))
hets: ./Persistence/DevGraph.hs:642:9-14: Maybe.fromJust: Nothing
```

It's the same problem again: A SymbolMapping is about to be created from `(Basic/Numbers version 1.0, 2, (op, "0"))` to `(Invoice_corrected, 47, (op, "0"))`, but the target symbol does not exist.



> using a faster db like mysql (the latter is probably easier?)

I'm not sure how this would work with MySQL. I only tested SQLite and PostgreSQL. The problem with PostgreSQL is the migration: We need Ontohub to do the migrations because of the enum types that are not fully supported by Persistent. It cannot migrate enum types. If we want to have Ontohub in the test suite, I can add it.

eugenk:
Edited the previous message: I forgot to post the name of the failing file "CaseStudies/Invoice_corrected.casl"

eugenk:
I tried adding MySQL support, but this fails at the migrations. Persistent-MySQL cannot handle class table inheritance, which we heavily use.

eugenk:
It's the same for PostgreSQL.

tillmo:
Here is a minimal example leading to the error. The problem seems to be the application of a generic unit, which leads to unnamed nodes. I have filed issue #1764.
```
library Invoice_corrected
spec Nat = sort s
spec ORDER = sort s
arch spec Warehouse =
units
  NatAlg : Nat;
  OrderFun : Nat -> ORDER;
result
  OrderFun[NatAlg]
end
```

eugenk:
I have found a workaround for the migration problem in 9efb165. Now, we can use PostgreSQL and MySQL in Travis, as implemented in c4e6f13. This also adds database configuration (yml) files.

tillmo:
rebasing to master should solve the problem with CaseStudies/Invoice_corrected.casl

eugenk:
I rebased on master. I also removed most of the files from the database test set. The failed file is now passing. Still, I need to run the complete test set on my computer to check if there are still some failures...

eugenk:
Rebasing solved the problem on the minimal example, but it still exists in the big InvoiceCorrected. It tries to create a SymbolMapping 
* from the symbol "0" (kind "op") from library "Basic/Numbers version 1.0", node "Basic/Numbers?Nat" (node id 2) 
* to the symbol "0" (kind "op") from library "Invoice_corrected", node "imports_union" (node id 47).
The latter cannot be found.

Here's my debug print:

```text
source: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
target: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
findSymbolInCache: (Basic/Numbers version 1.0,(2,"Basic/Numbers?Nat"),("op","\"0\""))
findSymbolInCache: (Invoice_corrected,(47,"imports_union"),("op","\"0\""))
```

tillmo:
I had to `sudo apt install libmysqlclient-dev` in order to make `make stack` succeed. This should be added to `README.md`.

tillmo:
for InvoiceCorrected, I filed #1769

tillmo:
The minimal example generating the problem is now
```
library Invoice_corrected
spec Nat = sort s
arch spec Warehouse =
units
  NatAlg : Nat;
  OrderFun : Nat -> Nat;
  T = OrderFun[NatAlg]
result
  OrderFun[T]
end
```
When merging in 1769-duplicate-node-names, the error `hets: Maybe.fromJust: Nothing` vanishes, but instead I get
```
hets: Persistence/DevGraph.hs:(472,17)-(502,32): Non-exhaustive patterns in case
```

mcodescu:
I have introduce a new DGOrigin in Static/DevGraph.hs, maybe this caused the error.

eugenk:
I have based this on the branch of 1769 in the branch [1691-dg_calculus_on_database-with-1769](https://github.com/spechub/Hets/tree/1691-dg_calculus_on_database-with-1769) and it still fails on Invoice_corrected when creating a SymbolMapping:

```text
source: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
target: CASL - Symbol {symName = 0, symbType = OpAsItemType (OpType {opKind = Total, opArgs = [], opRes = Nat})}
findSymbolInCache: (Basic/Numbers version 1.0,(2,"Basic/Numbers?Nat"),("op","\"0\""))
findSymbolInCache: (Invoice_corrected,(47,"imports_union"),("op","\"0\""))
hets: Maybe.fromJust: Nothing
```

tillmo:
rebasing on master now should solve the problems with the full `Invoice_corrected.casl` example.

eugenk:
Now it fails with
```text
hets: user error (Could not find target symbol for SymbolMapping:
symbol name: "0"
symbol kind: op
node name: 3_reduction
node ID: 73
library name: Examples/dodatek
```
on $HETS_LIB/Examples/dodatek.casl

tillmo:
Here is some minimal example, which however produces a different error:
```
library Examples/dodatek
spec ExpSemantics = sort Env
spec InstrSyntax = {}
spec InstrSemantics =
  InstrSyntax and ExpSemantics then
    free {
      op M : Env 
      }
end
```

eugenk:
This is because `let nodeLabel = getNodeLabel libEnv libName nodeId` in https://github.com/spechub/Hets/blob/601f4fa6a0de0b0a00e7b2fcd8484c1b42d0d0b1/Persistence/DevGraph.hs#L1074-L1081
assumes there is a nodeLabel to that nodeId and there is actually none. However, we need to at least mock a label at this point in order to save the OMS.

eugenk:
On another note, I think that we need to drop MySQL support. This database does not support indexes on text columns (only limited varchar). As a consequence, Loc/Ids cannot be saved as part of primary keys because they are too long (more than 255 characters for some examples in Hets-lib).

jelmd:
FWIW: Index capabilities depend on the storage engine in mysql. E.g.  according to https://dev.mysql.com/doc/refman/5.5/en/column-indexes.html the MyISAM storage engine supports FULLTEXT indexes (I guess the others not, because maintaining it is very expensive? ...).

tillmo:
OK, I have filed #1785

tillmo:
I filed #1792 which should solve the error with dodatek above

tillmo:
#1792 has been fixed, so please rebase on master

eugenk:
It's still the same error. even on the minimal example.

tillmo:
on dodatek.casl, I now get at least a different error:
```
hets: user error (./Persistence/DevGraph.hs:1172:3-8: Maybe.fromJust: Nothing
CallStack (from HasCallStack):
  error, called at ./Debug/Trace/LocationTH.hs:100:18 in loch-th-0.2.1-AlfMozdtyUtfKeBsBrgZJ:Debug.Trace.LocationTH)
```
which is not your original error (this should be fixed, #1792), but the error of the minimal example above (which is #1785 and still needs fixing).

tillmo:
OK, rebasing on master should now solve the problems.

eugenk:
DOL/align2.dol fails with
```text
hets: user error (Could not find target symbol for SymbolMapping:
symbol name: "http://www.hets.eu/ontology/unamed#Person_source"
symbol kind: Class
node name: align2?VAlignment_bridge
node ID: 4
library name: align2
```
and DOL/Alignment.dol fails with
```text
hets: user error (Could not find target symbol for SymbolMapping:
symbol name: "http://www.hets.eu/ontology/unamed#Woman"
symbol kind: Class
node name: Alignment?Alignment1_bridge
node ID: 4
library name: Alignment
```

eugenk:
These are the last two files from [test/hets-lib-testfiles](https://github.com/spechub/Hets/blob/0c1c431111e768803ecc7ac9c2725bec9b6787b4/test/hets-lib-testfiles) that fail with the database.

mcodescu:
I'll see if this can be fixed quick. Otherwise, should I expect more changes on your code?

eugenk:
I think there will not be many changes. I will remove the commit "DELETEME: Add debug prints", but that's about it.

eugenk:
Oh and maybe I will try to remove some duplication in Persistence.LogicGraph.

mcodescu:
I would guess that https://github.com/spechub/Hets/pull/1799 might be of help.

mcodescu:
Actually no: the problem is that the examples are written assuming that equality in correspondences will be encoded using morphisms, while the implementation uses a different assumption. See https://github.com/spechub/Hets/issues/1655, #1597.

tillmo:
my suggestion is to remove the last two files from test/hets-lib-testfiles, and reintroduce them once #1655 and #1597 have been solved.

mcodescu:
I totally agree.

eugenk:
Done. I also added a refactoring commit that removes some code duplication and improves the Persistence.LogicGraph module.

Still need to figure out how to make MySQL work.

eugenk:
The changes for MySQL will only affect a few lines of code (probably with some flags in the Schema.hs). Nothing that should stop you from reviewing.

eugenk:
There is a problem with symbols: After analysing HasCASL/TopologicalSpaces.het, we have 2932 (of 7485) symbols in the database that do not have a `symbol_kind`. This means that `symKind lid symbol` is an empty string:
https://github.com/spechub/Hets/blob/7039b16a64f748df7876d5f91e1a69ed95f5ff96/Persistence/DevGraph.hs#L777-L782
Hence, the `loc_id` may not be unique any more. One of these symbols without a kind is:
```
HasCASL/TopologicalSpaces.het//oms/RichTopologicalSpace/symbols//regularClosed
```
which is actually an `op`.
(Inside the double slash after "symbols" should be the kind, but it's empty)

mcodescu:
In https://github.com/spechub/Hets/blob/master/HasCASL/Logic_HasCASL.hs, there is no implementation for `symKind` in the instance declaration for class `Sentences`. It should be added, otherwise the default implementation returning the empty string is used.

tillmo:
I have filed #1800. Is HasCASL the only logic where this is missing?

mcodescu:
If I comment out the default implementation in `Logic.Logic`, I get a warning for missing implementation for the logics 

- CSMOF
- DMU
- Isabelle
- OMDoc
- QVTR
- Temporal
- HasCASL
- RelationalScheme
- CASL_DL
- COL
- CoCASL
- ConstraintCASL
- CSPCASL
- ExtModal
- FPL
- Modal
- VSE
- Adl
- CSL
- DFOL
- Framework
- FreeCAD
- HolLight
- LF
- Maude
- QBF

tillmo:
so then we would need an implementation for all of these in order to make the database connection work?

eugenk:
If they support more than one kind of symbol, yes. I can add a default implementation `"symbol"` to get rid of thew double slash, but we need to disambiguate the different kinds if there is more than one possible kind.

eugenk:
To do:

* [ ] Standalone database support: Create and find FileVersions properly. Don't create them again. Delete the old ones if needed. No versioning required.
* [ ] Dynamically load the database libraries. Do not require users to install MySQL and PostgreSQL.

Persistence/Schema/MappingType.hs
```diff
@@ -0,0 +1,65 @@
...
+
+data MappingType = LocalDef | LocalThmOpen | LocalThmProved
+                 | GlobalDef | GlobalThmOpen | GlobalThmProved
+                 | HidingDef
+                 | FreeDef | CofreeDef | NPFreeDef | MinimizeDef
+                 | HidingOpen | HidingProved
+                 | HidingFreeOpen | HidingCofreeOpen
+                 | HidingNPFreeOpen | HidingMinimizeOpen
+                 | HidingFreeProved | HidingCofreeProved
+                 | HidingNPFreeProved | HidingMinimizeProved
```
tillmo: these names should be changed: the "Hiding" needs to be removed from the last 8 constructors. See also https://github.com/ontohub/ontohub-backend/blob/add_graphql_types_for_document_data/app/graphql/types/mapping_type_enum.rb

eugenk: But they are constructed in LibEnv from HidingFreeOrCofreeThm and not FreeOrCofreeDefLink. See https://github.com/spechub/Hets/blob/9c217c0bf6d4b39f9f5eb721e19a9a3125998d60/Persistence/DevGraph.hs#L1114-L1127

tillmo:
`HidingFreeOrCofreeThm` means `hiding`, `free` or `cofree` theorem link.

eugenk:
I added indexes to the migrations and they increase the testing speed from 8 minutes to 5 minutes on my machine (same test set as on travis)

eugenk:
> Standalone database support: Create and find FileVersions properly. Don't create them again. Delete the old ones if needed. No versioning required.

Done.

eugenk:
Last thing to do (except for #1800 / https://github.com/spechub/Hets/pull/1752#issuecomment-346734486):
> Dynamically load the database libraries. Do not require users to install MySQL and PostgreSQL.

As a reminder for myself: https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/shared_libs.html#finding-shared-libraries-at-runtime

eugenk:
I asked a question on Stack Overflow about this problem: https://stackoverflow.com/questions/47680575/haskell-ghc-8-dynamically-load-import-module

tillmo:
@eugenk, now that #1800 is fixed, how urgent is #1808 for this pull request?

eugenk:
I would not use the database as long as #1808 is not fixed since it causes unique constraints to fail.

mcodescu:
I could work on #1808 but reviewing the code is quite complicated. Eugen, I would need some advice about what parts of your code should I focus on.

eugenk:
I will re-write the review guide. I think #1808 should be the priority now.

MGlauer:
Regarding the open problem 'Dynamically load the database libraries. Do not require users to install MySQL and PostgreSQL.': 
We could adapt the Makefile to build only those database modules whose packages are available and use the [`plugins`](https://hackage.haskell.org/package/plugins-1.5.6.0/docs/System-Plugins-Load.html) package to dynamically load them. I never used this library myself, but a dynamic plugin handling could be a vaulable extension for hets.

eugenk:
> adapt the Makefile to build only those database modules whose packages are available 

I'm sorry, I'm a little bit confused: Do you mean the Haskell packages (stack/stackage.org, cabal/hackage.org) or the operating-system packages (e.g. apt, pacman, brew)? The former are always available during build time because they are included in the Hets.cabal file. The problem exists during run time, not during build time. If you mean the latter: How can the build system know what operating systems packages will be available on the target machine?

Last time I checked the plugins package was incompatible to GHC 8, but maybe I checked it wrong. Now it seems to be compatible to GHC 8.0, but not 8.2. At least, it is not available directly in any stack snapshot with GHC 8, so there still may be compatibility problems.

eugenk:
I have tried it with the plugins package, but I failed because the compiled result cannot find the symbols from from the imported packages:

```
mysql -u root -e 'drop database if exists hets_test;'; mysql -u root -e 'create database hets_test;'; hets --quiet --output-types=db --database-config=Persistence/database_mysql.yml --database-subconfig=test --logic-graph
hets:
lookupSymbol failed in relocateSection (relocate external)
Persistence/MySQL.o: unknown symbol `_persistentzmmysqlzm2zi6zi2zi2zm96yQyay6azzPnBDy33elif_DatabaseziPersistziMySQL_withMySQLPool_closure'
hets: user error (resolvedObjs failed.)
```
crypric symbol name translated to human-readable format:
```
package: persistent-mysql-2.6.2.2
our module: Database.Persist.MySQL
symbol in package: withMySQLPool
type: closure
```

If someone can tell me how to compile the import into the object file or how to load the imported package, I'd be glad. Using [`loadPackage "persistent-mysql"`](https://hackage.haskell.org/package/plugins-1.5.6.0/docs/System-Plugins-Load.html#v:loadPackage) in the first line of https://github.com/spechub/Hets/blob/9a45d483040a027805bbba9b5c4789a821b97fa8/Persistence/DynamicDatabaseConnection.hs#L48-L54 does not help at all.

eugenk:
If nobody can help with this, I suggest to drop MySQL support (only keep SQLite and PostgreSQL). MySQL is causing problems with String-column widths anyway. Also, the dynamic loading problem seems to only exist for MySQL.

eugenk:
I think this is ready for review now. I deactivated MySQL for the reasons mentioned in the previous posts. I also rebased on master and updated the review guide. Installing Ontohub is not needed any more.

mcodescu:
I have started with Readme then went through LogicGraph, DevGraph, Diagnosis and FileVersion, all in Persistence. The ideas described in Readme are very solid. This is quite different from the regular Hets updates, also refactoring the way the specifications are stored internally. I will surely need to adapt to this. But the code is well written, function names carefully chosen and this helps a lot. Since all checks have passed, I see no reason not to merge this into the main branch. @eugenk: really nice work!

tillmo:
When running hets without Postgresql installed, I get
```
./hets: error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory
```
@MGlauer will have a more detailed look into `plugins` and report about this.

eugenk:
Todo:
Use C-preprocessor statements to only load PostgreSQL support for hets-server.

eugenk:
I disabled PostgreSQL for hets-desktop and added it as a dependency to the debian package of hets-server. This should be finished now.

tillmo:
I cannot compile:
```
make stack
ghc-pkg: cannot find package glade
ghc-pkg: cannot find package glade
ghc-pkg: cannot find package uni-uDrawGraph
/usr/local/bin/stack build --install-ghc --only-dependencies 
/home/till/projects/Hets-server/_dependencies/persistent/persistent-mysql/: getDirectoryContents:openDirStream: does not exist (No such file or directory)
Makefile:41: die Regel für Ziel „stack“ scheiterte
make: *** [stack] Fehler 1
```
Probably the following line should be deleted from `stack.yml`:
```
  - location: './_dependencies/persistent/persistent-mysql'
```

tillmo:
now I get
```
make stack
ghc-pkg: cannot find package glade
ghc-pkg: cannot find package glade
ghc-pkg: cannot find package uni-uDrawGraph
/usr/local/bin/stack build --install-ghc --only-dependencies 
/home/till/projects/Hets-server/_dependencies/persistent/persistent-postgresql/: getDirectoryContents:openDirStream: does not exist (No such file or directory)
Makefile:41: die Regel für Ziel „stack“ scheiterte
make: *** [stack] Fehler 1
```

eugenk:
> I cannot compile:

You don't seem to have initialised the git submodule.

> Probably the following line should be deleted from stack.yml:

No, it is needed because of changes in a shared `persistent-*` package.

tillmo:
you are right, I forgot the submodule, sorry. Now I get:
```
make restack
[...]
Progress: 22/23make
persistent-sqlite-2.6.3: copy/register
Completed 23 action(s).

--  While building package persistent-mysql-2.6.2.1 using:
      /home/till/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2 --builddir=.stack-work/dist/x86_64-linux/Cabal-1.24.2.0 build --ghc-options " -ddump-hi -ddump-to-file"
    Process exited with code: ExitFailure 1
    Logs have been written to: /home/till/projects/Hets-server/.stack-work/logs/persistent-mysql-2.6.2.1.log

    Configuring persistent-mysql-2.6.2.1...
    Building persistent-mysql-2.6.2.1...
    Preprocessing library persistent-mysql-2.6.2.1...
    [1 of 1] Compiling Database.Persist.MySQL ( Database/Persist/MySQL.hs, .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/Database/Persist/MySQL.o )
    
    /tmp/stack2124/persistent-mysql-2.6.2.1/Database/Persist/MySQL.hs:1189:17: error:
        • Could not deduce (BackendCompatible r0 backend)
            arising from a use of ‘projectBackend’
          from the context: (backend ~ PersistEntityBackend record,
                             BackendCompatible SqlBackend backend,
                             PersistEntity record,
                             MonadIO m)
            bound by the type signature for:
                       insertManyOnDuplicateKeyUpdate :: (backend
                                                          ~
                                                          PersistEntityBackend record,
                                                          BackendCompatible SqlBackend backend,
                                                          PersistEntity record, MonadIO m) =>
                                                         [record]
                                                         -> [SomeField record]
                                                         -> [Update record]
                                                         -> ReaderT backend m ()
            at Database/Persist/MySQL.hs:(1176,1)-(1186,27)
          The type variable ‘r0’ is ambiguous
          Relevant bindings include
            insertManyOnDuplicateKeyUpdate :: [record]
                                              -> [SomeField record]
                                              -> [Update record]
                                              -> ReaderT backend m ()
              (bound at Database/Persist/MySQL.hs:1187:1)
          These potential instances exist:
            instance BackendCompatible SqlBackend SqlBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
            instance BackendCompatible SqlBackend SqlReadBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
            instance BackendCompatible SqlBackend SqlWriteBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
        • In the first argument of ‘withReaderT’, namely ‘projectBackend’
          In the first argument of ‘(.)’, namely ‘withReaderT projectBackend’
          In the expression: withReaderT projectBackend . uncurry rawExecute
    
    /tmp/stack2124/persistent-mysql-2.6.2.1/Database/Persist/MySQL.hs:1190:15: error:
        • Could not deduce (BackendCompatible SqlBackend r0)
            arising from a use of ‘rawExecute’
          from the context: (backend ~ PersistEntityBackend record,
                             BackendCompatible SqlBackend backend,
                             PersistEntity record,
                             MonadIO m)
            bound by the type signature for:
                       insertManyOnDuplicateKeyUpdate :: (backend
                                                          ~
                                                          PersistEntityBackend record,
                                                          BackendCompatible SqlBackend backend,
                                                          PersistEntity record, MonadIO m) =>
                                                         [record]
                                                         -> [SomeField record]
                                                         -> [Update record]
                                                         -> ReaderT backend m ()
            at Database/Persist/MySQL.hs:(1176,1)-(1186,27)
          The type variable ‘r0’ is ambiguous
          These potential instances exist:
            instance BackendCompatible SqlBackend SqlBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
            instance BackendCompatible SqlBackend SqlReadBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
            instance BackendCompatible SqlBackend SqlWriteBackend
              -- Defined in ‘persistent-2.7.2:Database.Persist.Sql.Orphan.PersistStore’
        • In the first argument of ‘uncurry’, namely ‘rawExecute’
          In the second argument of ‘(.)’, namely ‘uncurry rawExecute’
          In the expression: withReaderT projectBackend . uncurry rawExecute
Makefile:44: die Regel für Ziel „restack“ scheiterte
make: *** [restack] Fehler 1
```

eugenk:
I changed one commit. You seem to have got the state before my last force-push. The bad commit was online for only a short amount of time.

eugenk:
> Just checked: artful+ (and thus next LTS) is comming with PG 9.6. Are you sure, that hets will work with this as well as 10+?

Yes. I have used it with PSQL 9.6 and 10.1 and it runs just fine.

> Having a *-dev dependency is usually wrong, since server environments are per default not developer environments, i.e. have only the runtime stuff installed. So why is the libpq-dev required?

You're right. I just copied over the dependencies from the readme. I removed the dev dependency with 2c01f52ea

tillmo:
Thanks, now it works. For `hets_server`, I just needed to install `libpq5` and `postgresql`, not `libpq-dev`.
Could you also please copy your usage explanations above into `doc/UserGuide.tex`?

tillmo:
P.S. note that `libpq-dev` has `libpq5` as a dependency, and one only needs that dependency.

eugenk:
I didn't even need to install libpq5. Tested this with a fresh ubuntu 16.04 vm.

eugenk:
Done in b0224f9.
Also rebased on master

doc/UserGuide.tex
```diff
@@ -1379,6 +1380,16 @@ \section{Reading, Writing and Formatting}
...
 
+The \texttt{db} option will save the development graph to a database.
+This can be either an SQLite database or a PostgreSQL database (only available
+in the \texttt{hets-server} package). For an SQLite database, you need to pass
+the parameter \texttt{--database-file=FILEPATH} to specify the database file to
+save to. If the file does not yet exist, \Hets will create it.
+For a PostgreSQL database, you need to pass the
+\texttt{--database-config=FILEPATH} and \texttt{--database-subconfig=KEY}
+parameters that point to a Ruby-on-Rails-compatible \texttt{database.yml}
+configuration file.
```
tillmo: could you please include a sample database.yml file, or point to the sample file on github?

eugenk: Done with 6b71baeef. I linked it to the master branch, hoping that it will remain there.

tillmo:
I have tried the workflow described at the top of this page in a VM: both `libpq5` and `postgresql` are needed.

tillmo:
Eventually! Congratulations!

PR #1762, opened 2017-10-16
names for nodes in unit specs

mcodescu:
With
```
spec sp1 =
  sort s
end
unit spec usp1 = sp1 -> sp1
end

spec sp2 = sort t
end

spec sp = sp1 and sp2

unit spec usp2 = sp1 * sp2 -> sp
```
I get 
graph

### DIFF
--- a/Static/AnalysisArchitecture.hs
+++ b/Static/AnalysisArchitecture.hs
@@ -19,8 +19,6 @@ module Static.AnalysisArchitecture
     , anaRefSpec
     ) where
 
-import Debug.Trace
-
 import Driver.Options
 
 import Logic.Logic
@@ -1019,8 +1017,7 @@ anaArgSpecs lgraph libEnv ln dg opts eo usName args = let
        l <- lookupLogic "anaArgSpecs " (currentLogic lgraph) lgraph
        let sp = item argSpec
            xName = makeName $ usName{abbrevPath = abbrevPath usName ++ show x}
-       (argSpec', argSig, dg') <- trace ("xName:" ++ show xName ++ 
-                                         "\nsp:" ++ show sp)$
+       (argSpec', argSig, dg') <- 
            anaSpec False False -- don't optimize the node out 
               lgraph libEnv ln aDG (EmptyNode l) 
               (xName{extIndex = x + 1})
--- a/Static/AnalysisArchitecture.hs
+++ b/Static/AnalysisArchitecture.hs
@@ -166,13 +166,16 @@ alreadyDefinedUnit u =
     "Unit " ++ iriToStringUnsecure u ++ " already declared/defined"
 
 -- | Create a node that represents a union of signatures
-nodeSigUnion :: LogicGraph -> DGraph -> [MaybeNode] -> DGOrigin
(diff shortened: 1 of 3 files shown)


tillmo:
good. Could you please use a nonzero `extIndex` for the new `NodeName`s? Then the generated node names are not displayed by default.

mcodescu:
Done but I now get a strange behavior. With the `extIndex` 1 or x+1 for the `NodeName xName` in the method `Static.AnalysisArchitecture.anaArgSpecs`, the nodes for parameters of a parameterized unit spec are no longer added to the development graph (!?!). If the `extIndex` is 0, as set by the method `Static.DGUtils.makeName`, the nodes are there and have different names. Do you have any idea what goes wrong?

tillmo:
These nodes are optimized away. Look at lines 519-526 of `Static.AnalysisStructured`. Suggestion: add a flag indicating that this optimization shall not take place.

mcodescu:
Yes, I just found that too, I'll follow your suggestion.

Static/AnalysisArchitecture.hs
```diff
@@ -927,7 +953,8 @@ lambda expressions, like you do in the following -}
    do
      -- beh will be ignored for now
      (_rsig@(BranchRefSig _ (usig, _)), dg', asp') <-
-           anaUnitSpec lgraph libEnv ln dg opts eo nsig nP uspec
+           anaUnitSpec lgraph libEnv ln dg opts eo rn nsig nP uspec 
+                                       -- TODO: is rn fine above?
```
tillmo: could you tell why it should / should not be fine?

mcodescu: I am thinking we want 1. to follow some convention about the names of the nodes (with this argument we will get R_args0, ... R_argsn, R_res as node names for the unit specification being refined, which is weird) and 2. to make sure that we don't get the same node names twice (but now I think this is not possible, because the name R of the refinement is unique).

mcodescu: Using rn here is bad, we get duplicate names. I'm looking into that, please do not merge until I fix it.

mcodescu:
Now for the refinement `refinement R = sp1 -> sp1 refined to sp1 -> sp2` the nodes are called
`R_source_arg0`, `R_source_res` and respectively `R_target_arg0` and `R_target_res`, with some underscores and a number at the end. Thus we make sure that the names of the nodes are unique. Got into some issues with modifying the IRIs in such a way that they get displayed right and fixed them by writing a new function `addSuffixToIRI` which adds a string to an IRI such that the string is displayed by `iriToStringShort`.

tillmo:
interesting, do your IRI concerns  relate to #1596?

mcodescu:
No: the name of the refinement R is an IRI with `abbrevPath` the name of the dol file and with `abbrevQuery` ?R while the name of the unit spec usp1 is an IRI with `abbrevPath `usp1 and `abbrevQuery` empty. Then the function `iriToStringShort` displays only the tail of `abbrevQuery` if it's not empty and otherwise both the `abbrevPath` and the `abbrevQuery`. I couldn't tell you why this is so, I just adapted to this.

PR #1771, opened 2017-11-01
1769 duplicate node names

mcodescu:


Common/IRI.hs
```diff
@@ -921,3 +922,9 @@ mergeCurie c i =
 
 deleteQuery :: IRI -> IRI
 deleteQuery i = i { abbrevQuery = "" }
+
+addSuffixToIRI :: String -> IRI -> IRI
+addSuffixToIRI s i = if not $ null $ abbrevQuery i 
+                   then i{abbrevQuery = abbrevQuery i ++ s}
+                  else  
+                        i{abbrevPath  = abbrevPath i ++ s}
```
tillmo: please have "then" and "else" equally indented, and please delete the line break after "else"

Static/DevGraph.hs
```diff
@@ -1591,3 +1592,11 @@ duplicateDefEdges :: DGraph -> [Edge]
...
   filter (not . isSingle) . group . map (\ (s, t, _) -> (s, t))
   . filter (liftE isDefEdge) . labEdgesDG
+
+ensureUniqueNames :: DGraph -> IRI -> Int -> NodeName
+ensureUniqueNames dg anIRI n = 
+ let allNames = map (getName . dgn_name . snd) $  labNodesDG dg
+     iri' = addSuffixToIRI (show n) anIRI
+ in if iri' `elem` allNames 
+     then ensureUniqueNames dg anIRI (n + 1) 
+     else makeName iri'
```
tillmo: please file an issue that this should be refactored in such a way that `allNames` is always cached as a component of the DG.

Common/IRI.hs
mcodescu: Done

Static/DevGraph.hs
mcodescu: see #1772

Issue #1796, opened 2017-11-16
export id of nodes and edges in DGXML / DGjson


tillmo:
DG datastructure: `Static.DevGraph`, graph datastructure: https://hackage.haskell.org/package/fgl-5.6.0.0/docs/Data-Graph-Inductive-Graph.html
Files to change: function `lnode` in `Static.ToJson` and `lnode` in `Static.ToXml`.

PR #1797, opened 2017-11-16
Add id's to Nodes

ysengrimm:
see #1796 

Edges already have the id stored in ```linkid```.
Nodes will have the id stored in an ```id``` field.


Static/ToXml.hs
```diff
@@ -98,13 +98,14 @@ prettySymbol :: (GetRange a, Pretty a) => GlobalAnnos -> a -> Element
...
 lnode :: HetcatsOpts -> GlobalAnnos -> LibEnv -> LNode DGNodeLab -> Element
-lnode opts ga lenv (_, lbl) =
+lnode opts ga lenv (nodeId, lbl) =
   let nm = dgn_name lbl
       (spn, xp) = case reverse $ xpath nm of
           ElemName s : t -> (s, showXPath t)
           l -> ("?", showXPath l)
   in add_attrs (mkNameAttr (showName nm)
     : rangeAttrs (srcRange nm)
+    ++ [mkAttr "id" (show $ nodeId)]
```
tillmo: the $ is superfluous here. Or you write `mkAttr "id" $ show nodeId`, then a $ makes sense.

ysengrimm:
I implemented the source and target id's for both json and xml.  
Edges now have `id_source` and `id_target`.

PR #1799, opened 2017-11-22
made unique names for the nodes in W-diagram of an alignment

mcodescu:
I now use the function `addSuffixToIRI` for the nodes in the W-diagram of an alignment A, such that they are now called A_source, A_target and A_bridge, and these names appear in the graph too. Note that it is not necessary to call `ensureUniqueNames`, as there can't be two alignments with same name in a library.


mcodescu:
While not essential for #1752, this is still needed because at least on my machine, the master branch does not display the suffixes of the nodes.

PR #1802, opened 2017-11-30
1801 redesign webgui

### LINKED ISSUE #1801: redesign webgui
I suggest to redesign the existing web gui, because it would enable potential users to test Hets without any barriers. 

here is a suggestion for a design of the page. 
bildschirmfoto 2017-11-27 um 11 05 26

eugenk:
This shall fix #1801. It is not a good redesign, but it should be much more useful than before. Given this has taken two whole days to accomplish because the code is hard to understand, I wouldn't want to make more big changes at this point.

screen shot 2017-11-30 at 17 27 18
screen shot 2017-11-30 at 17 27 25
screen shot 2017-11-30 at 17 27 29

Or for a bigger file:
screen shot 2017-11-30 at 17 29 10


tillmo:
DOLiator should be replaced by "Hets, the DOLiator". After all, we should not neglect Hets.

tillmo:
The button "List all commands as XML" should be renamed into "Information about the restful interface".

Moreover, in the development graph window, the "commands" button should be renamed into "further commands", and the first command (auto) should be provided with a separate button "graph: automated proofs".

tillmo:
I have added some more meaningful examples

eugenk:
There is still one page that I need to redesign: I didn't see that there is a proving interface.

eugenk:
> The button "List all commands as XML" should be renamed into "Information about the restful interface".

This is *not* the restful interface. This is rather a stateful, session-based web interface. There is another web interface that more or less implements REST. I suggest to name the button "Information about the Web Interface".

eugenk:
I think this is done for the most part. There are still pages that are in the old design, but the old code is so hard to understand that I don't want to change it unless you really need it for the conference.

tillmo:
The button previously called "List all commands as XML" is not about the web interface either. So it should be deleted.

tillmo:
Also, the xupdate - impact - submit buttons on the development graph page should be removed.

tillmo:
None of my comments under "looks good! Some comments:" has been implemented.

tillmo:
in a sense, this _is_ the RESTful interface, namely a `GET /` on the RESTful interface.

tillmo:
so there should be a link to the description of the RESTful interface at https://github.com/spechub/Hets/wiki/RESTful-Interface

eugenk:
> * selecting examples from Hets-lib is a bit awkward, because if you click on a folder, you have to click again on "More examples" before you see its contents. Also, would it be possible to place the "more examples" below the minimal ones? Then one does not need to scroll so much

Done.

> * why is there a selection of the output format already on the start page? It seems to have no effect

I don't know. It was there from the beginning. I hadn't tested the effect on the start page. This is removed now.

 * generated PDFs are corrupt, although the look like PDFs if examining the contents

I can imagine that this bug also exists on master.  I haven't changed that code.

> Also, the xupdate - impact - submit buttons on the development graph page should be removed.

Done

> The button previously called "List all commands as XML" is not about the web interface either. So it should be deleted.

Done

eugenk:
> * could you preserve linkebreaks of prover output by using appropriate HTML formatting options?

I will try this, but I don't have much time. I have other plans for today and there are only a few minutes left until I need to leave.

tillmo:
good, for now I have approved the pull request even without those last changes (the line breaks and the link to the wiki page explaining the RESTful interface).

eugenk:
I have redesigned the proof results page. I hope this is what you wanted to have. Now I need to leave. If you agree with these changes, you can merge them without me.

tillmo:
good will do this

tillmo:
yes, great!

Issue #1809, opened 2017-12-09
in several logics, improve symbols and symbl kinds


mcodescu:
In Modal, ExtModal, ConstraintCASL, CoCASL, COL, CASL_DL, VSE symbols are now just CASL symbols. They should be extended to cover the symbols from the extension part of CASL signatures.

Also, for some logics, the type of symbols is `()`. This should also be improved where meaningful

Issue #1808, opened 2017-12-07
implement symbol kind for various logics


tillmo:
see #1800 as a role model. Implement it for these logics (but note that some of them are built over CASL and therefore can re-use what is there to CASL):
- CSMOF
- DMU
- Isabelle
- OMDoc
- QVTR
- Temporal
- RelationalScheme
- CASL_DL
- COL
- CoCASL
- ConstraintCASL
- CSPCASL
- ExtModal
- FPL
- Modal
- VSE
- Adl
- CSL
- DFOL
- Framework
- FreeCAD
- HolLight
- LF
- Maude
- QBF

tillmo:
@mcodescu: this one is important fpr #1752. Could we try to distribute the work among us somehow? Note that for each logic, it should only be little effort. For some logics, even one default kind might be sufficient.

mcodescu:
I'll start and tell you where I run into problems, or logics I don't understand.

mcodescu:
See the branch https://github.com/spechub/Hets/tree/1808-symbol-kinds. Issues:
 - in some places, I have used emptyKind, but I think this should rather be the
default in Logic.Logic. Agreed?
 - only one kind in LF. Should I call it ops or leave it to default emptyKind?
 - in HolLight we have () as type for symbols, and no instance of class Sentences???
 - in CSL the Symbol type seems to have not enough information about the symbols
as the kind is not stored. Should we change it?
 - in VSE there is a comment about how the type of symbols should be extended...
 - in Modal symbols are now just CASL symbols. Should be corrected?
 - same for ExtModal
 - same for ConstraintCASL
 - same for CoCASL
 - same for COL
 - same for CASL_DL

tillmo:
* yes, `emptyKind' should be the default in Logic.Logic
 * in LF, I suggest `consts`
 * HolLight is indeed incomplete. @sternk, could you please help?
 * CSL: yes, should be changed
 * VSE: where is the comment?
 * Modal etc.: yes, there are new types of symbols, like modalities, of flexible ops

mcodescu:
- I have made "emptyKind" a default. 
- changed to consts
- VSE: in Logic_VSE.hs, `morphisms and symbols need to be extended, too`

sternk:
@tillmo - From what I remember HolLight is a very shallow logic node that
is only meant to facilitate reading dumps directly into the DG which are
then fed into the HolLight2Isabelle comorphism. What are the currently
[...]

PR #1810, opened 2017-12-09
1808 symbol kinds

mcodescu:
Solves #1808, but see #1809

### DIFF
--- a/Adl/Sign.hs
+++ b/Adl/Sign.hs
@@ -60,9 +60,9 @@ data Symbol
 data SymbolKind = ConK | RelK 
   deriving (Eq, Ord, Show, Typeable, Data)
 
-symType :: Symbol -> SymbolKind
-symType (Con _) = ConK
-symType (Rel _) = RelK
+sym_kind :: Symbol -> SymbolKind
+sym_kind (Con _) = ConK
+sym_kind (Rel _) = RelK
 
 
 instance Pretty SymbolKind where 
--- a/Adl/Sign.hs
+++ b/Adl/Sign.hs
@@ -57,6 +57,18 @@ data Symbol
   | Rel Relation
     deriving (Eq, Ord, Show, Typeable, Data)
 
+data SymbolKind = ConK | RelK 
+  deriving (Eq, Ord, Show, Typeable, Data)
+
+symType :: Symbol -> SymbolKind
+symType (Con _) = ConK
+symType (Rel _) = RelK
+
+
+instance Pretty SymbolKind where 
+ pretty ConK = text "concept"
+ pretty RelK = text "relation" 
+
 instance GetRange Symbol where
   getRange s = case s of
     Rel r -> getRange r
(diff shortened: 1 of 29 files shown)


Adl/Sign.hs
```diff
@@ -57,6 +57,18 @@ data Symbol
   | Rel Relation
     deriving (Eq, Ord, Show, Typeable, Data)
 
+data SymbolKind = ConK | RelK 
+  deriving (Eq, Ord, Show, Typeable, Data)
+
+symType :: Symbol -> SymbolKind
+symType (Con _) = ConK
+symType (Rel _) = RelK
```
tillmo: I think `symKind` is a better name for this function, because of the result type (and symbol types and kinds are distinguished e.g. in CASL).

Logic/Logic.hs
```diff
@@ -380,7 +380,7 @@ class (Language lid, Category sign morphism, Ord sentence,
       fullSymName l = show . sym_name l
       -- | a logic dependent kind of a symbol
       symKind :: lid -> symbol -> String
-      symKind _ _ = ""
+      symKind _ _ = "emptyKind"
```
tillmo: I rather suggest `defaultKind` here.

Maude/Symbol.hs
```diff
@@ -68,6 +70,28 @@ type SymbolSet = Set Symbol
 type SymbolMap = Map Symbol Symbol
 type SymbolRel = Rel Symbol
 
+data SymbolKind = SortK | KindK | LablK | OpK 
+        deriving (Show, Read, Ord, Eq, Typeable)
+
+-- | the kind of a symbol
+
+symType :: Symbol -> SymbolKind
```
tillmo: also here, use `symKind`

RelationalScheme/Sign.hs
```diff
@@ -63,6 +65,17 @@ data RSSymbol = STable Id |     -- id of a table
                     RSIsKey     -- is it a key?
                 deriving (Eq, Ord, Show, Typeable, Data)
 
+data RSSymbolKind = STableK | SColumnK 
+  deriving (Eq, Ord, Show, Typeable, Data)
+
+symType :: RSSymbol -> RSSymbolKind
```
tillmo: also here, use `smyKind`

Adl/Sign.hs
mcodescu: To do this, the function in Logic.Logic should be renamed to `sym_kind`, otherwise I get an error about ambiguities when defining `symKind Adl` in terms of `symKind`. Agreed?

Logic/Logic.hs
mcodescu: done

Adl/Sign.hs
mcodescu: Actually it took less replacements to change `symType` with `sym_kind` for Adl, Maude and RelScheme.

tillmo: good

Issue #1806, opened 2017-12-07
In DGXML and DGjson, add attribute indicating whether a node is internal


tillmo:
use `Static.DgUtils.isInternal`

PR #1818, opened 2017-12-19
Add `internal` field to nodes

ysengrimm:
see #1806 

Adds an `internal` field to both DGJson and DGXML.


tillmo:
please merge (the approval is now 14 days old)

PR #1827, opened 2018-01-12
Remove Persistent-fork and use Esqueleto

eugenk:
This removes the fork of Persistent that allowed to use Postgres enumeration types in database columns (see https://github.com/yesodweb/persistent/issues/264 for the currently unresolved issue) and adds Esqueleto.

We want to use Esqueleto anyway (in future pull requests) because of its feature to create `SELECT JOIN` SQL-statements. This library, however, is not compatible to the fork of Persistent.

If the new `select` calls look unusual to you, please see the `select` calls in #1829. The you should be able to see why it's necessary to `return` the arguments of the function pass to `from`. These arguments specify which tables to use in the SQL `SELECT` statement. Sometimes, not all of these tables need to be included in the result set.

### DIFF
--- a/Hets.cabal
+++ b/Hets.cabal
@@ -134,6 +134,7 @@ Executable hets
       , http-types >= 0.6 && <= 1.0
       , text >= 0.5 && < 1.3
       , case-insensitive
+      , esqueleto >= 2.5.3
       , file-embed >= 0.0.10
       , persistent >= 2.7.0
       , persistent-template >= 2.5.2
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "_dependencies/persistent"]
-	path = _dependencies/persistent
-	url = https://github.com/eugenk/persistent.git
-	branch = add_support_for_postgresqls_enumerated_types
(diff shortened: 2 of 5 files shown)


Hets.cabal
```diff
@@ -134,6 +134,7 @@ Executable hets
       , http-types >= 0.6 && <= 1.0
       , text >= 0.5 && < 1.3
       , case-insensitive
+      , esqueleto >= 2.5.3
```
jelmd: Is 2.4.1 sufficient as well?

eugenk: 2.4.1 is incompatible with persistent 2.7 and (more importantly) GHC 8.

jelmd: Bah. Is there somewhere a note about it? A changelog or similar stuff doesn't seem to exist it.

eugenk: I simply tried to downgrade to esqueleto 2.4.1 and compile Hets. Resolving the dependencies failed.

eugenk: There is no other note about compatibility. Unfortunately, they don't maintain changelogs.

jelmd: Ah ok, thought you have other info channels.
Did also some digging, but it seems that 2.4.1 needs persistent-"old" which probably breaks the stuff you already made. Anyway, at least bionic will have 2.5.3 and thus I'll add this to xenial PPA, too.

PR #1830, opened 2018-01-16
simplify version checking in makefiles

jelmd:
Please test on non-Ubuntu platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@
 
 # This GNU Makefile will compile the hets system and provides also
 # targets for test programs during implementation phases.
+#
+# If the following environment variables are set, their values get passed to
+# the corresponding tool as is: GHC_PKG_FLAGS (ghc-pkg), GHC_FLAGS (ghc).
 
 include var.mk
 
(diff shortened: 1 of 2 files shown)


eugenk:
Compiles just fine on macOS 10.13.2 without ksh93.

eugenk:
On another note (since you are changing the Makefiles and): 
Are there any supported Ubuntus left that don't have a warp3 (but only a warp1) package?

jelmd:
Wrt. warp1 + ubuntu. Not sure, but at least trusty. Anyway, doesn't really matter because one does not have to use the system packages ;-) .

jelmd:
Wrt. MacOS: good - wasn't sure (my Mac is dead, accu bumped to 3x size). IIRC MacOSX (at least tiger .. *capitan) shipped with ksh93. Not anymore?

eugenk:
macOS is shipped with the executable `ksh`:
```
❯ /bin/ksh --version
  version         sh (AT&T Research) 93u+ 2012-08-01
```
but not with with an executable named`ksh93`.

It is, however, always shipped with `/bin/bash`.

jelmd:
Ahh ok, thanx.

BTW: Just saw, that the commit already contains the GHC_PKG_FLAGS. So will add another change, which honors GHC_FLAGS, so that one may use some flags on-the-fly.

jelmd:
Anything else?

eugenk:
Still need to test this on arch linux. Will do this tomorrow.

eugenk:
Compiles fine on Arch Linux, too.

> doesn't really matter because one does not have to use the system packages ;-)

Isn't this required in a PPA?

jelmd:
PPA requirements: either system packages or packages from the PPA in use.
However, if one maintains its own "Haskell Platform" (e.g. because needs to avoid certification process for all "new modules" ...) it may contain, whatever is needed/sufficient for his primary projects.

PR #1834, opened 2018-01-17
make mysql support optional

jelmd:
persistent-mysql & friends do not exists on Ubuntu (neither xenial nor bionic+) - perhaps because of license concerns. Making mysql optional (disabled by default as currently is hardcoded anyway) allows one to build on bionic+ with system modules ...

### DIFF
--- a/Hets.cabal
+++ b/Hets.cabal
@@ -61,6 +61,9 @@ flag haxml
 flag httpclient
   description: use native http-client implementation instead of wget
 
+flag mysql
+  description: enable persistent support for MySQL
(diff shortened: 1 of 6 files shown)


jelmd:
I can try, but in a separate branch, because it is a little bit more complicated, probably takes more time to get it right and would be easier to review. Wrt. Ubuntu this one is more important.

eugenk:
Alright. Then, this is good to go after the conflict has been resolved.

jelmd:
Done.

PR #1836, opened 2018-02-26
Reasoning with database

eugenk:
This should add database support for reasoning. It changes the REST API wrt. Proving/Consistency-Checking.

Please review this carefully because there might be some typos when changing the state (moving through libEnv1, libEnv2, libEnv3 and so on while proving many conjectures). Please also check if there are any logical mistakes like not using the correct gTheory for proving/consistency-checking or not passing the set of all sentences (even imported ones) to the SInE premise selection.

# REST API
* POST `/prove/coded-iri/command-list` is for invoking a prover on a proof goal. It accepts the following request body
    ```text
    {
      "format": "json", // or "xml" or "db" //optional
      "goals": [ // required
        {
          "node": "Nat__E1", // required
          "conjecture": "ga_comm___+__", // optional, if left out, every conjecture will be proved on its own with this configuration
          "reasonerConfiguration": { // required
            "timeLimit": 10, // seconds // required
            "reasoner": "EProver" // optional
          },
          "premiseSelection": { // optional
            "kind": "manual", // or "sine" // required
            "manualPremises": ["Ax1", "Ax2"], // required if kind is "manual"
            "sineDepthLimit": 5, // optional (default: unlimited)
            "sineTolerance": 1.7, // optional (default: 1.0)
            "sinePremiseNumberLimit": 50 // optional (default: unlimited)
          },
          "useTheorems": true, // use previously proven goals as axioms // optional (default: true)
          "printDetails": true, // print details of the proof in xml/json output // optional (default: true)
          "printProof": true // print the prover output in xml/json output // optional (default: true)
        }
      ]
    }
    ```
    Example call with curl:
    ```text
    curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "Nat__E1", "conjecture": "ga_comm___+__", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "EProver"}, "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/consistency-check/$(urlencode $HETS_LIB/Basic/Numbers.casl)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
    ```
* POST `/consistency-check/coded-iri/command-list` is for invoking a consistency-checker on a node. It accepts the following request body
    ```text
    {
      "format": "json", // or "xml" or "db" //optional
      "goals": [ // required
        {
          "node": "Nat__E1", // required
          "reasonerConfiguration": { // required
            "timeLimit": 10, // seconds // required
            "reasoner": "darwin" // optional
          },
          "useTheorems": true, // use previously proven goals as axioms // optional (default: true)
        }
      ]
    }
    ```
    Example call with curl:
    ```text
    curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "Nat__E1", "reasonerConfiguration": {"timeLimit": 10}}]}' "localhost:8000/prove/$(urlencode $HETS_LIB/Basic/Numbers.casl)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
    ```

# To Do
* [x] Refactor `buildReasoningCache`. See https://github.com/spechub/Hets/pull/1836#discussion_r170591991. Done in 0d3de413a.
* [x] Remove the `parameterize` call of logic translations. See https://github.com/spechub/Hets/pull/1836#issuecomment-368876543 for the required escaping and a few comments above that one for the general problem. Done in a different way in 957a8be74.
* [ ] Completely change the Sentence/LocIdBase structure. See https://github.com/spechub/Hets/pull/1836#issuecomment-369156065 for a possible solution.
  * [ ] Adjust the database schema
  * [ ] Save OMSLocIdSentence objects for all imported and non-imported sentences
  * [ ] Fetch the correct sentence from the database during reasoning.


tillmo:
when compiling, I get a number of "missing module" warnings. I am just curious why missing modules do not cause errors. Have you employed some dynamic loading of modules at run-time?
```
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnConjectureType
  Probable cause: out-of-date interface files
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnReasoningAttemptType
  Probable cause: out-of-date interface files
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnConjectureType
  Probable cause: out-of-date interface files
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnReasoningAttemptType
  Probable cause: out-of-date interface files
[837 of 866] Compiling Driver.WriteFn   ( Driver/WriteFn.hs, Driver/WriteFn.o ) [Persistence.LogicGraph changed]
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnConjectureType
  Probable cause: out-of-date interface files
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnReasoningAttemptType
  Probable cause: out-of-date interface files
[839 of 866] Compiling Persistence.Reasoning ( Persistence/Reasoning.hs, Persistence/Reasoning.o )
[862 of 866] Compiling PGIP.Server      ( PGIP/Server.hs, PGIP/Server.o )
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnConjectureType
  Probable cause: out-of-date interface files
WARNING in hptSomeThingsBelowUs
  missing module Persistence.Schema.ReasoningStatusOnReasoningAttemptType
  Probable cause: out-of-date interface files
```

eugenk:
This is because of some left-overs that GHC hasn't detected `make clean` should make that go away.

tillmo:
good, thanks. Now I have tried your example call and get:
```
curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "Nat__E1", "conjecture": "ga_comm___+__", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "EProver"}, "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/consistency-check/$(urlencode $HETS_LIB/Basic/Numbers.casl)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
*** Error:
user error (no cons checker found)
```
This is strange because 1) I have tried to prove, not to cons-check, 2) I have cons checkers available when using the old Hets GUI.

eugenk:
Is `EProver` a valid cons-checker?

tillmo:
no

tillmo:
OK, will try with "prove"

eugenk:
Then it is filtered away in
https://github.com/spechub/Hets/blob/c7bf6ef68b469d5cc673e24f18d2f866bf7704e1/PGIP/Server.hs#L1839-L1847
which is quite old Hets-code (just moved out into a separate function)

tillmo:
OK, I managed to prove a goal and found the prover output in table `reasoner_outputs`. I also managed to check consistency of an OMS. However, I did not find the output of the consistency checker in the DB.

PGIP/GraphQL/Result/Action.hs
```diff
@@ -0,0 +1,9 @@
+{-# LANGUAGE CPP, TypeFamilies, DeriveDataTypeable #-}
+
+module PGIP.GraphQL.Result.Action where
+
+import Data.Data
+
+data Action = Action { evaluationState :: String
+                     , message :: Maybe String
+                     } deriving (Show, Typeable, Data)
```
tillmo: why is this datatype called `Action`? It seems to be more a datatype for storing results of some action?

eugenk: This might be extended to the action datatype that Tim described in his master's thesis. It can get an ETA and more information about the action in the future.

eugenk:
> I did not find the output of the consistency checker in the DB.

That's because Hets does not capture the output in the old function `consNode`.
https://github.com/spechub/Hets/blob/c7bf6ef68b469d5cc673e24f18d2f866bf7704e1/PGIP/Server.hs#L1821-L1837

PGIP/GraphQL/Result/Action.hs
tillmo: good, then please add a corresponding comment.

tillmo:
OK, but could you then please create an issue for that?

PGIP/Server.hs
```diff
@@ -553,7 +543,15 @@ parseRESTful
               cmdOptList
             newOpts = foldl makeOpts opts $ mapMaybe (`lookup` optionFlags)
               optFlags
-        in if elem newIde newRESTIdes && all (`elem` globalCommands) cmdList
+            validReasoningParams =
+              if elem newIde ["prove", "consistency-check"]
+              then case reasoningParametersE of
+                     Left _ -> False
+                     Right _ -> True
+              else True
```
tillmo: shorter:
`not (validReasoningParams = elem newIde ["prove", "consistency-check"]) || Data.Either.isRight reasoningParametersE `
or even
`Data.Either.isRight reasoningParametersE <= (validReasoningParams = elem newIde ["prove", "consistency-check"])`
Here, `<=` is read as "if", i.e. reverse "implies". See also https://mail.haskell.org/pipermail/libraries/2016-January/026565.html

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
+    failOnLefts :: ReasoningCacheE -> ResultT IO ()
+    failOnLefts reasoningCache =
+      let lefts_ = filter isLeft reasoningCache
+      in  if null lefts_
+          then return ()
+          else fail $ unlines $ map (\ (Left message_) -> message_) lefts_
+
+    isLeft :: Either a b -> Bool
+    isLeft (Left _) = True
+    isLeft _ = False
```
tillmo: this function is already provided by Data.Either

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
+
+    unRight :: Either a b -> b
+    unRight (Right x) = x
+    unRight _ = error "PGIP.Server.unRight received a Left."
```
tillmo: shorter: `unRight = fromRight $ error "PGIP.Server.unRight received a Left."`
Perhaps you could make the error message a bit more informative?

eugenk:
> OK, but could you then please create an issue for that?

I _think_ I added it in 055599943.

tillmo:
yes, works, great!

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
+  return (libEnv', nodesAndProofResults)
+  where
+    useDatabase :: Bool
+    useDatabase = format reasoningParameters == Just "db"
+
+    failOnLefts :: ReasoningCacheE -> ResultT IO ()
+    failOnLefts reasoningCache =
+      let lefts_ = filter isLeft reasoningCache
+      in  if null lefts_
+          then return ()
```
tillmo: shorter: `if null $ lefts reasoningCache`

PGIP/Server.hs
eugenk: I cannot use `fromRight`. It could not be found even though I have imported `Data.Either`.

PGIP/Server.hs
tillmo: aha, or better: `let lefts_ = lefts reasoningCache`

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
-  -> [String] -> [String] -> ResultT IO (LibEnv, [ProofResult])
+        (gConsChecker, comorphism) : _ -> return (gConsChecker, comorphism)
+
+reasonREST :: HetcatsOpts -> LibEnv -> LibName -> DGraph -> ProverMode
+           -> String -> ReasoningParameters
+           -> ResultT IO (LibEnv, [(String, [ProofResult])])
+reasonREST opts libEnv libName dGraph_ proverMode location reasoningParameters = do
+  reasoningCacheE <- liftIO $ buildReasoningCache
+  failOnLefts reasoningCacheE
+  let reasoningCache1 = map unRight $ filter (not . isLeft) reasoningCacheE
```
tillmo: shorter: `let reasoningCache1 = rights reasoningCacheE`
(you already know that there are only `Right` elements)

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
+                                 , rceComorphism = comorphism
+                                 }
+                               ) goalNames ++ reasoningCacheE2)
+                  )
+                  reasoningCacheE1
+                  goalConfigByNode
+      )
+      [] $
+      groupBy (\ a b -> ReasoningParameters.node a == ReasoningParameters.node b) $
+        ReasoningParameters.goals reasoningParameters
```
tillmo: I find it hard to review `buildReasoningCache`. Could you please give names to the anonymous functions (even if this means passing more arguments around), and also to the `groupby` term at the end?

eugenk: Will do. Could take me some time because I'm working on a different task right now.

PGIP/Server.hs
```diff
@@ -1806,27 +1822,224 @@ consNode :: LibEnv -> LibName -> DGraph -> (Int, DGNodeLab)
...
+      (libEnvAcc2, proofResults2) <-
+        foldM
+          (\ (libEnvAcc3, proofResults3) reasoningCacheGoal -> do
+            -- update state
+            let gTheoryM = maybeResult $ getGlobalTheory nodeLabel
+            gTheory_ <- case gTheoryM of
+              Nothing ->
+                fail ("Cannot compute global theory of: "
+                      ++ showName (dgn_name nodeLabel) ++ "\n")
+              Just gTheory_ -> return gTheory_
```
tillmo: this is a repetition of lines 1898ff. Could be relegated into a small function.

tillmo: this is a repetition of lines 1898ff. Could be relegated into a small function.

jelmd:
BTW: In Issue ??? you asked for DB "driver" separation as done for MySQL. This would at least in theory allow a "no DB at all" configuration. So related questions are: Is that a supported config? If so, is the new code prepared for it (i.e. relevant parts enclosed in e.g. #ifdef USE_DB ... #endif )?

eugenk:
It is compatible to a no-db configuration.

eugenk:
The relevant code is only called if Hets has been called with database-related commandline-options.

tillmo:
I have a question about `logic_translations`. When I prove a simple first-order goal with a TPTP prover like SPASS, in the json I get
```
 "used_translation": "CASL2SoftFOL",
```
In the DB, I get
```
INSERT INTO "logic_translations" VALUES(1,'id-casl-condEq-casl2softfol');
```
How can a comorphism be reconstructed from `id-casl-condEq-casl2softfol'`? After all, this comorphism has two constituents, but the name uses three dashes.

tillmo:
The last question is relevant when I want to specify a translation (possibly a composition of logic mappings) that should be used for proving. Now my question is: how can I do this? I tried
```
till@mark:~/projects/Hets$ curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "sp", "conjecture": d=c", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "SPASS"}, "translation": "casl2softfol", "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/prove/$(urlencode /home/till/temp/prove.dol)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
*** Error:
user error (No matching translation or prover found for sp
)
till@mark:~/projects/Hets$ curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "sp", "conjecture": d=c", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "SPASS"}, "translation": "CASL2SOFTFOL", "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/prove/$(urlencode /home/till/temp/prove.dol)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
*** Error:
user error (No matching translation or prover found for sp
)
```
By the way, the error message would be more informative if the translation or prover that has not been found were included into the message.

tillmo:
It seems to be that only `Persistence.LogicGraph.findLogicMappingBySlug` can find a logic mapping, given a string. However, this function assumes that the logic mapping is already stored in the database. What if the user specifies a logic mapping (with several constituents) not present in the database, to be used for proving?

tillmo:
There seems to be problem with imported axioms. Consider
```
spec sp =
  sort s
  ops a,b,c:s
  . a=b %(a=b)%
  . b=c %(b=c)%
end

spec sp1 =
sp then
  . a=c %(a=c)% %implied
end
```
I get
```
curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "sp1", "conjecture": "a=c", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "SPASS"}, "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/prove/$(urlencode /home/till/temp/prove3.dol)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"
*** Error:
Persistence.Reasoning.saveToDatabase: Could not find Sentence file:///home/till/temp/prove3.dol//oms/sp1/sentences/a=b 
```
although the sentence is there (but for OMS sp, not sp1!)
```
INSERT INTO "sentences" VALUES(41,36,12,'a=b','. a = b');
INSERT INTO "oms" VALUES(36,35,18,80,2,NULL,6,NULL,NULL,NULL,NULL,'dg_basic_spec',9,11,'sp','sp','',0,0,0,'Open');
INSERT INTO "oms" VALUES(43,35,19,80,2,NULL,7,NULL,NULL,NULL,NULL,'dg_basic_spec',10,14,'sp1','sp1','',0,0,0,'Open');
C
```

Persistence/Reasoning.hs
```diff
@@ -0,0 +1,677 @@
...
+        case premiseSelection $ rceGoalConfig reasoningCacheGoal of
+          Nothing -> return (Nothing, 0, NoResult)
+          Just premiseSelectionParameters -> do
+            let premiseSelectionKindV =
+                  getPremiseSelectionKind premiseSelectionParameters
+            let gTheory = rceGTheory reasoningCacheGoal
+            let goalName = fromJust $ rceGoalNameM reasoningCacheGoal
+            performPremiseSelection opts gTheory goalName
+              premiseSelectionParameters premiseSelectionKindV
+
```
tillmo: I would omit this blank line, or else also add one before the `return`below.

eugenk:
About the logic translations:

I could remove the `parameterize` call when creating the slug (essentially making it `id` on the name of the translation). However, there is a character used in translation names that is not allowed in URIs, namely `>` (from `->`).

eugenk:
About the problem with imported axioms:

This might be a problem in `Persistence.DevGraph`. I think that imported sentences are not stored in the database.

On another note: It is possible to use `=` in a sentence name (`a=b`)? This could be a problem with Loc/Ids of sentences because they are URIs and there are restrictions to characters that can be used there. `=` should be a character that is not allowed.

tillmo:
Concerning logic translations: yes, and maybe just escape `>`. But my question also was: What if the user specifies a logic mapping (with several constituents) not present in the database, to be used for proving?

tillmo:
> I think that imported sentences are not stored in the database.
indeed, they are stored, but only for the imported oms, not for the importing oms. That could be repaired in several ways:
1. store such sentences also for the importing oms (meaning that the sentence is store twice, or in general multiple times)
1. store each sentence once, but use a more complicated algorithm when resolving a loc/id (basically, look at all the directly or indirectly imported oms as well) 
1. instead of a sentence having an oms_id, use a join table between sentences and oms, such that a sentence can belong to several oms.

eugenk:
> What if the user specifies a logic mapping (with several constituents) not present in the database, to be used for proving?

Shouldn't be a problem because the logic mapping is found by the non-database-part of Hets. The slug is only used to find it in the database to avoid duplicates.

The logic mapping itself is managed by old Hets code and the database entry is only used by Ontohub to display which logic mapping has been used.


> they are stored, but only for the imported oms, not for the importing oms. That could be repaired in several ways

Does Hets guarantee that no two sentences in an OMS have the same name? Consider your Document in https://github.com/spechub/Hets/pull/1836#issuecomment-368540998: Could `sp1` define a sentence with the name `a=b`? If so, we have a problem in general with the Loc/Ids. Otherwise:

The sentence itself has a fixed `loc_id` in the database which is the only thing that allows for a performant lookup when an object with a Loc/Id is queried by the user. This means:
1. Is possible, but also takes up a lot more space in the database. However, space is cheaper than computation time of a lookup.
1. Not possible because the `loc_id` includes the OMS's `loc_id` as a prefix and it is fixed in the database. Changing this would overcomplicate the query
1. Same as 2.

tillmo:
> Does Hets guarantee that no two sentences in an OMS have the same name? Consider your Document in #1836 (comment): Could sp1 define a sentence with the name a=b? If so, we have a problem in general with the Loc/Ids.

currently, this is allowed. But we could forbid it.

I think we can still store each sentence only once, using a join table linking oms and sentences. Additionally, sentences have an oms_id that determines the main oms of the sentences, which is the oms where the sentence directly occurs. This oms is also used for generating the loc/id of the sentence.

eugenk:
We don't build the loc_id during a lookup. We build it when saving an object to the database. See [Schema/LocIdBase](https://github.com/ontohub/ontohub-models/blob/784aeaaddf425aac0e0bb711817923e33e9d900f/db/migrate/20160902111740_initial_schema.rb#L275) and [Schema/Sentence](https://github.com/ontohub/ontohub-models/blob/784aeaaddf425aac0e0bb711817923e33e9d900f/db/migrate/20160902111740_initial_schema.rb#L530-L531) and [Query for a Sentence](https://github.com/ontohub/ontohub-backend/blob/e8ce595ac5d3cb7226e5a7a061fd0ac762c4181e/app/graphql/types/git/commit_type.rb#L171). The whole loc_id is used in a `SELECT * FROM table INNER JOIN loc_id_bases WHERE loc_id = <loc_id>` statement.

We would need to change the whole structure of Loc/Ids of sentences:

```text
   ----------------------------------------------
   |                                            |
   |                                            |
-------       --------------------      ------------------
| OMS |  ---- | OMSLocIdSentence | ---- | Sentence       |
-------       --------------------      ------------------
              | oms_id           |      | primary_oms_id |
              | sentence_id      |      ------------------
              | loc_id           |
              --------------------
```
This means, that we can't inherit `Sentence < LocIdBase` any more (meaning that we need to copy over everything from `LocIdBase` except for the `loc_id` field) and we need to query for the sentence by `loc_id` differently. Could become complicated in the Ruby-part...

eugenk:
I'm not reluctant to changing this. I'm just saying that it could take a while to complete this task.

eugenk:
To summarise, here is what I need to do:

* [ ] Refactor `buildReasoningCache`. See https://github.com/spechub/Hets/pull/1836#discussion_r170591991
* [ ] Remove the `parameterize` call of logic translations. See https://github.com/spechub/Hets/pull/1836#issuecomment-368876543 for the required escaping and a few comments above that one for the general problem.
* [ ] Completely change the Sentence/LocIdBase structure. See https://github.com/spechub/Hets/pull/1836#issuecomment-369156065 for a possible solution.
  * [ ] Adjust the database schema
  * [ ] Save OMSLocIdSentence objects for all imported and non-imported sentences
  * [ ] Fetch the correct sentence from the database during reasoning.

I copied this checklist to the initial post of this pull request.
I may start doing this on Sunday.

tillmo:
I see. In this case, I suggest that `OMSLocIdSentence` becomes a subclass of `LocIdBase` and it is renamed into `Sentence`, and `Sentence` into `SentenceOrigin`. All associations that previously went into `Sentence` should now go into the new `Sentence`, except perhaps that to `Symbol`.

eugenk:
Like this?
```text
-------------
| LocIdBase |
-------------
| loc_id    |
-------------
     ^
     |
     |-----------------|
     |                 |
     |                 |
  ------- 1    * ---------------------- *     1 ----------------------------------------
  | OMS | ------ | Sentence (new)     | ------- | SentenceOrigin (previously Sentence) |
  -------        ----------------------         ----------------------------------------
                 | oms_id             |         | primary_oms_id                       |
                 | sentence_origin_id |         | name                                 |
                 ----------------------         | text                                 |
                          ^                     ----------------------------------------
                          |
                          |
                          |----------------
                          |               |
                     ----------     --------------
                     |  Axiom |     | Conjecture |
                     ----------     --------------
                                          |
                                         ...
                                  (Reasoning-Models)
```

tillmo:
ye, exactly.

tillmo:
I just realised that in `Sentence (new)`, we also need the `text`, because it can happen that the sentence is renamed along an import. This means that  `Sentence (new)` comes very close to  `Sentence`, and the refactoring will merely be adding `SentenceOrigin`, and filling more sentences into `Sentence (new)`.

eugenk:
So `SentenceOrigin` is only a reference to the OMS defining the sentence. How do I get all sentences along with the Node that defines it? What about recursively imported sentences?

eugenk:
What do you think of this?
```text
-------------
| LocIdBase |
-------------
| loc_id    |
-------------
     ^
     |
     |-----------------|
     |                 |
     |                 |
  ------- 1    * ------------------------
  | OMS | ------ | Sentence (new)       |
  -------        ------------------------
                 | oms_id               |
                 | original_sentence_id |
                 | name                 |
                 | text                 |
                 ------------------------
                          ^
                          |
                          |
                          |----------------
                          |               |
                     ----------     --------------
                     |  Axiom |     | Conjecture |
                     ----------     --------------
                                          |
                                         ...
                                  (Reasoning-Models)
```

where `original_sentence_id` is `null` if it is not imported.

eugenk:
* I refactored `buildReasoningCache` in 0d3de413a. I hope this makes the function easier to understand.
* Instead of removing the `parameterize` call, I added a name to a LogicTranslation in 957a8be74. The slug is only supposed to be used by Ontohub (in URLs) to display the details of the translation. The name is Hets-internal and is used to specify a translation when performing an action like proving.

tillmo:
> So SentenceOrigin is only a name and a reference to the OMS defining the sentence. 

yes

> How do I get all sentences along with the Node that defines it? What about recursively imported sentences?

Static.ComputeTheory.computeTheory gives you all the sentences and the (recursively) imported sentences of a node. However, I think without the defining node - you would need to add that (suggestion: extend the function computeTheory with that, rename it, and let computeTheory call the new function and remove the extra info.

tillmo:
yes, your new database structure is even better.

eugenk:
Right, Static.ComputeTheory.computeTheory gives me a G_theory with all sentences, but there is no information about the origin of a sentence. My first thought was to simply subtract the local sentences (`dgn_theory nodeLabel`) from the computed theory (`Static.ComputeTheory.computeTheory dGraph nodeId`), but I think this won't work: How do I compare the sentences to decide whether or not to remove them? I cannot compare them by name because they could have been renamed in the import.

tillmo:
moreover, there can be recursive imports.
By the way, I think that 'original_sentence_id` is a nice-to-have, but you could have a first version without it. Then computeTheory suffices.

eugenk:
I think if I don't add this now, then it will never be added

tillmo:
OK, you have a point...

eugenk:
bcc3bef20 saves all sentences in the database, but it does not change the schema. I'm not sure, though, how to tackle the problem with a sentence's origin.

eugenk:
I don't think that I can solve the remaining todos without some infrastructure in Hets that allows me to easily get the source OMS+Sentence of a Sentence. This information should be saved during static analysis, I guess. This should be done in another issue (or two issues: a) adding the information during analysis and b) saving a corresponding link in the database).

On another note, the example in https://github.com/spechub/Hets/pull/1836#issuecomment-368540998 does not fail any more. 

Are there any other things left to do?

tillmo:
OK, I filed two issues. However, what is still missing: after a successful proof attempt, I want to know a) which reasoner and b) which axioms have been used. For the above mentioned example, I cannot find this in the database.  Note that in `reasoner_configurations`, the `configured_reasoner_id` is `NULL`.

eugenk:
> I want to know a) which reasoner 

This is already done. This is what I see in psql:
```
hets_development=# SELECT * FROM reasoning_attempts;
 id |     kind     | action_id | reasoner_configuration_id | used_reasoner_id | used_logic_translation_id | time_taken
----+--------------+-----------+---------------------------+------------------+---------------------------+------------
  1 | ProofAttempt |         6 |                         1 |               20 |                         1 |          0
  2 | ProofAttempt |         7 |                         2 |               20 |                         1 |          0
(2 rows)
```
In id 1, I configured SPASS directly and in id 2, I did not state which reasoner to use.

I used the following commands:
```
psql -U postgres -c 'drop database hets_development;'; \
psql -U postgres -c 'create database hets_development;'; \
hets --output-types=db --database-config=Persistence/database_postgresql.yml --database-subconfig=development --logic-graph; \
hets --output-types=db --database-config=Persistence/database_postgresql.yml --database-subconfig=development --hets-libdirs=$HETS_LIB /tmp/test_hets_reasoning.dol; \
hets --output-types=db --database-config=Persistence/database_postgresql.yml --database-subconfig=development --hets-libdirs=$HETS_LIB -X

# And in another terminal:

curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "sp1", "conjecture": "a=c", "reasonerConfiguration": {"timeLimit": 10, "reasoner": "SPASS"}, "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/prove/$(urlencode /tmp/test_hets_reasoning.dol)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"

# and

curl -X POST -H "Content-Type: application/json" -d '{"format": "db", "goals": [{"node": "sp1", "conjecture": "a=c", "reasonerConfiguration": {"timeLimit": 10}, "premiseSelection": {"kind": "sine", "sineDepthLimit": 2, "sineTolerance": 1, "sinePremiseNumberLimit": 50}}]}' "localhost:8000/prove/$(urlencode /tmp/test_hets_reasoning.dol)/auto/full-theories/full-signatures?hets-libdirs=$(urlencode $HETS_LIB)"

```

> and b) which axioms have been used. 

This is left out completely - I have forgotten about it. The database schema needs to be adjusted.

tillmo:
> >   I want to know a) which reasoner

>This is already done. This is what I see in psql:

Thanks for clarifying, I had overlooked that.

eugenk:
Do we want to save which generated axioms have been used? In the old Ontohub, we only save which axioms _from before the translation_ have been used.

eugenk:
If only the used sentences from before the translation should be saved, 16f270a70 should do it.

tillmo:
> Do we want to save which generated axioms have been used? In the old Ontohub, we only save which axioms from before the translation have been used.

good question. I think "from before" suffices, but I cannot exclude the possibility that we will discover some situation where we need "after". But for now, "from before" suffices.

tillmo:
Note that the origin of sentences is now available, see #1837

eugenk:
I still would like to fix the origin in another pull request. I'm not sure if I should do something about travis not finishing for PostgreSQL. It will be difficult though, because recreating the cache and compiling consumes almost the whole time quota.

tillmo:
OK, then I suggest to merge this now.

luisenriqueramos1977:
Hi, I am having the issue: 
*** Error: no matching translation or prover found, when I try to prove an specification where an axiom is imported like the example commented by  **tillmo on 26 Feb 2018**.
I wonder if there is a solution for such issue?.

Luis Ramos

PR #1839, opened 2018-03-27
origin of sentences

mcodescu:
I have added a `Maybe SenOrigin` in the labelling of sentences. If `senOrigin` of a sentence is `Nothing`, then the sentence was introduced in the current node, otherwise it is imported. 

I was already half-way through with adding the `LibEnv` as an argument when needed, so I have not taken into account @tillmo's suggestion to add it to the `GTheory`. If still needed, maybe we can do this in a different issue.

I have tested the code on very simple files, and it seems to me that it does what it should.


mcodescu:
One more thing, `dGraphName` is an `Id` instead of a `LibName` because importing `Common.LibName` causes an import cycle.

tillmo:
@eugenk needs an IRI or a LocId, so maybe you could directly use that?

mcodescu:
I have switched to IRI. If you would prefer LocId, is there a method already that builds one from a LibName?

tillmo:
I think IRI is OK, because it is used in Hets at many places, wheras LocId is only used in `Persistence`. @eugenk, what do you think?

eugenk:
I think an IRI is not be best option. I need something that I can easily use to query for a sentence. The best thing that pops into mind is a tuple (libName, nodeId, sentenceName).

mcodescu:
Yes, we have a tuple now: 
  - name of the library, which is an IRI
  - node id, as Node
 - sentence name, as String
As I said, the reason why the library name is an IRI and not a LibName is that it would cause a cycle of imports. I guess one could search for the LibName whose IRI is the one stored in the tuple.

tillmo:
@mcodescu, your code looks good. But I found one problem. I have checked in a branch `print_sen_origin` that also outputs origins of sentences (with a list of sentence names and their origins, at the end of a theory). With
```
hets -o th -v2 Basic/RelationsAndOrders.casl
```
Hets writes (among others) a file `Basic/RelationsAndOrders_AntisymmetricRelation.th`, and this contains at the end (in the mentioned list, which here is a singleton)
```
antisym Nothing
```
meaning that the axiom `antisym` is stored without origin.

mcodescu:
See my first comment in the pull request: if senOrigin of a sentence is Nothing, then the sentence was introduced in the current node, otherwise it is imported.

tillmo:
OK. Would it be difficult to store the origin also in case of local axioms?

mcodescu:
This gives me an easy way of identifying local axioms and thus only their origin gets changed when importing from a node, and not the origin of the axioms that are imported in that node. I'm sure there could be other ways, e.g. adding a `isLocal` flag. I was thinking this way of representing origin of sentences reduces redundancy. But if you prefer we can think of a better solution.

tillmo:
OK, fine with me, but then please add a comment explaining this (probably best when the datatype is introduced).

PR #1849, opened 2018-05-11
first try to see how this would work

mcodescu:
Not intended to merge, but pull request shows diff better.

### DIFF
--- a/Logic/Logic.hs
+++ b/Logic/Logic.hs
@@ -387,6 +387,9 @@ class (Language lid, Category sign morphism, Ord sentence,
       -- | combine two symbols into another one
       pair_symbols :: lid -> symbol -> symbol -> Result symbol
       pair_symbols lid _ _ = error $ "pair_symbols nyi for logic " ++ show lid
+      -- | rename a symbol 
+      rename_symbol :: lid -> symbol -> Id -> symbol
+      rename_symbol lid _ _ = error $ "symbol renaming nyi for logic " ++ show lid
 
 -- | makes a singleton list from the given value
 singletonList :: a -> [a]
--- a/OWL2/AS.hs
+++ b/OWL2/AS.hs
@@ -354,6 +354,9 @@ mkEntity = Entity Nothing
 mkEntityLbl :: String -> EntityType -> IRI -> Entity
 mkEntityLbl = Entity . Just
 
+renEntity :: Entity -> Id -> Entity
+renEntity e i = e{cutIRI = idToIRI i}
+
 instance Ord Entity where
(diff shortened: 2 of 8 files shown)


tillmo:
I have tried this extensively with 29 DOL files that BKB has provided. It mostly works, with two exceptions. The first one is: for compound ids, it is not possible to omit the symbol kind. Consider e.g.
```
logic OWL

spec S[Class: X] =
 Class: D SubClassOf: X
end

spec SC = S[C[X]]
end
```

mcodescu:
I made your example work (not an elegant solution, but it goes through). If other of the test files involving compound identifiers do not work, please let me know.

tillmo:
Good, now more things go through!
However, the following does not work (probably due to the nesting):
```
logic OWL
ontology A [ObjectProperty: p] [Class: X] = {}
ontology B [ObjectProperty: p] = {}
ontology C [ObjectProperty: p] [Class: D] = B[A[p][Y[D]]]
```
leads to
```
*** Error:
file does not exist: D.env
no input found for: D
```

tillmo:
Another problem: Let `A.dol` contain
```
logic OWL
ontology A = Class: C 
```
and `B.dol` contain
```
from A get A
logic OWL
ontology P[A] = {}
ontology B[A] = P[A]
```
Then at the left-hand side of `= `,  `A` correctly is interpreted as an ontology, but a the right-hand side, it is interpreted as a new class.

tillmo:
Another point: compound IDs with multiple components do not work. Consider
```
logic OWL
ontology O [Class: A] = {}
ontology P = O[C[D,E]]
```
I get
```
*** Error:
file:///home/till/temp/cc.dol:3.19:
unexpected ','
```

mcodescu:
The last issue requires a modification of the parser for structured specifications.

tillmo:
I see. Luckily, the last issue is the least important one.

mcodescu:
I modified a bit the test for the second issue, to make it more visible: 
```
from A get A
logic OWL
ontology P[A] = {Class: D SubClassOf: C}
ontology B[A] = P[A]

ontology Test = B[AClass]
```
This goes through now.

mcodescu:
The first two issues seem to be fixed now, please test and let me know if there are problems.

tillmo:
many thanks, goes through as expected!
However, there is a problem in Algebra_II:
```
file:///home/till/CASL/Hets-lib/Basic/Algebra_II.casl:216.35-216.41:
*** Error:
try to instantiate a formal parameter with more than one symbol with a single-symbol spec:Group
```

tillmo:
once #1899 has been closed, this one can go

PR #1853, opened 2018-06-06
fix package dependencies and switch over to x86_64 arch, only.

jelmd:


jelmd:
PR is now finished, ready for review. It fixes the debian/control file to show the correct dependencies required to build related debian alias ubuntu packages.
It also fixes the travis-ci stuff, so that now PRs do not require 2-3+ re-starts to get finished, saves several hours of useless computation per PR CI-run and makes DB test execution reliable.

utils/travis-helper.sh
```diff
@@ -0,0 +1,475 @@
...
+# TRAVIS_BUILD_NUMBER=2016
+# TRAVIS_BUILD_DIR=/home/travis/build/spechub/Hets
+# TRAVIS_BRANCH=master
+# TRAVIS_EVENT_TYPE=pull_request
+# TRAVIS_PULL_REQUEST=1853
+# TRAVIS_PULL_REQUEST_BRANCH=dependency_fix
+# TRAVIS_JOB_ID=396668281
+# TRAVIS_JOB_NUMBER=2016.1
+
+[[ -z "${PUSH_URL}" ]] && PUSH_URL='http://theo.cs.ovgu.de/cgi-bin/push'
```
ysengrimm: Can this be abused to upload (and download) arbitrary data to the server in any way? Also, will the external cache automatically cleaned after a set amount of time, or is this a manual task?

.travis.yml
```diff
@@ -1,102 +1,173 @@
...
+        - ${TH} myPrepareBuildDir
+        - cd ${HETS_BUILD_DIR}
+        # Compile Hets
+        - make stack && touch ${STACK_ROOT}/bin/stack.done
+        - stack exec -- ghc -V
+
+
+    # .2
+    - stage: Build
+      if: type != push AND env(SKIP_BUILD) != 1
```
ysengrimm: Why are the build and test stages not run when code is pushed?

utils/travis-helper.sh
jelmd: Basically yes. However, to prevent DoS, the upload directory is a separate filesystem. For upload the parameters have to match certain patterns (little bit "security by obscurity" ;-)), and uploads/downloads are restricted to IKS intern and travis-CI sites, only. Last but not least, directory indexing is disabled, so one needs to know, what to download.

So with some effort and proper timing, cache poisoning is possible. However, I think it is unlikely. If it becomes a problem, stuff can be probably enhanced by using Travis-CI encryption stuff (haven't had the time to study this stuff).

.travis.yml
jelmd: When the PR gets pushed (e.g. to the master), builds and tests have been run already, so no need to run/delay it again for 50+ min. Also there is usually no value in caching build results, because PRs should always do a complete, clean build to avoid usage of old artifacts from previous builds.

However, if this works as I expect wrt. the shallow travis-ci docs, PRs will benefit from building Stack on push, because all PRs can use the same cached Stack build. If one thinks, stack should be rebuild, one can manually delete the caches via travis-ci web ui, and all starts from zero.

ysengrimm: Ok this makes sense 👍.

ysengrimm:
Should I go ahead and merge this @tillmo ?

tillmo:
yes please

PR #1878, opened 2019-04-15
Default to tptp

mcodescu:


tillmo:
when using the above example UserManual/Chapter3.casl, I still get comorphism `id_CASL.Prenex=;CASL2SoftFOL` selected.

mcodescu:
Could you please tell me where is the order for comorphisms paths available for 'Select comorphism path:' for proving is set? I could not find anything that would give priority to `suleCFOL2SoftFOL` over `suleCFOL2TPTP`, and modifying the order in which the logics SoftFOL and TPTP or the translations to TPTP and SoftFOL appear in the lists of logics, respectively comorphisms does not bring anything, so it must be set somewhere. I also find it strange that for UserManual/Chapter3.casl we get in this list the same translation to TPTP twice.

tillmo:
The problem was that there were two instances of the SPASS prover with the same name. I have pushed a fix.

mcodescu:
It works now. Do you have to approve it?

mcodescu:
On my machine I have EProver and eprover available for UserManual/Chaper3.casl. The first comes from TPTP and the other from SoftFOL. The question is why SPASS_SoftFOL is not available. Also, should we also rename eprover to EProver_SoftFOL?

tillmo:
> It works now. Do you have to approve it?

you have to update the branch (see the button here on github)

tillmo:
> On my machine I have EProver and eprover available for UserManual/Chaper3.casl. The first comes from TPTP and the other from SoftFOL. The question is why SPASS_SoftFOL is not available. Also, should we also rename eprover to EProver_SoftFOL?

yes, we should rename it. 
Actually, SPASS_SoftFOL is listed with `hets -z`, but not available in the "prove" menu. I think the old comorphism should be re-added somehow to `KnownProvers.hs`, but not as the default one.

PR #1885, opened 2019-05-12
differentiate two instances of Pellet prover

tillmo:


mcodescu:
Please add a comment in the code explaining what PelletEL is. Also there seems to be a problem with the checks: `The command "sudo -E apt-add-repository -y "ppa:hets/hets"" failed and exited with 1 during .`

jelmd:
Wrt. failed apt-add-repository: sometimes travis-ci seems to have connection or resource problems (overloaded?) and than obscure errors appear. If not sure, just restart the job e.g. 10 min later (click on the **Details** link on the PR page) and check, whether the same error happens again ...

tillmo:
I have removed the PelletEL prover, because it is just Pellet restricted to a sublogic of OWL2, which does not make sense. Please approve.

PR #1886, opened 2019-05-15
1823 webgui translation

tillmo:
Adds a "translate" button to the DOLiator web GUI (available via `hets -X`)


tillmo:
hm, can you try again? For me, it works quite fast...

mcodescu:
Now I notice a speed-up: couple of seconds, maybe 5, for the first time the list is computed and perhaps 2-3 for a new selection. Both now and at first attempt I was on the same connection, its speed should not be an issue.

PR #1888, opened 2019-05-18
1887 het to dol

tillmo:
change extension .het to .dol


ExtModal/Tries/Makefile
```diff
@@ -2,9 +2,9 @@
 
 .PHONY: check
 
-check: hets
-	../../hets -v2 -o pp.het,th Try?.het
```
mcodescu: it should stay `hets -v2 -o pp.dol`

Fpl/test/Makefile
```diff
@@ -2,9 +2,9 @@
 
 .PHONY: check
 
-check: hets
-	../../hets -v2 -o pp.het,th specs.het
-	../../hets -v2 -o pp.het,th specs.pp.het
+check:.dols
```
mcodescu: should stay `hets -v2 -o pp.dol` and perhaps `specs.dol`

mcodescu:
Otherwise there are some files in `utils` that use `.het`.

tillmo:
> For backwards compatibility reasons, would it be possible to also support `.het` in the same way as `.dol`?

We still support .het as input format, but only pp.dol (not pp.het) as output format (supporting both would be more complicated)

tillmo:
> Otherwise there are some files in `utils` that use `.het`.

I have now renamed all `.het` files into `.dol`. I could not find any `.het` file in `utils`.

tillmo:
OK, tests go through, please review again

Issue #1905, opened 2019-07-18
How to translate CASL into TPTP File?


gibso:
How can i translate CASL files into TPTP files? 
As casl input file i am using the minimal example file from hets. The content is the following:
```
spec Strict_Partial_Order =
sort Elem
pred __<__ : Elem * Elem
forall x, y, z:Elem
. not x < x %(irreflexive)%
. x < y /\ y < z => x < z %(transitive)%
%% the following is logically entailed
. x < y => not y < x %(asymmetric)% %implied
end
```

now I am using the following command:
```
hets -o tptp example.casl
```

that gives me the following output:
```
Analyzing file file:///data/casl/example.casl as library example
Analyzing spec Strict_Partial_Order
expected RDF theory for: ./example_Strict_Partial_Order.tptp
```
the exit code of that command is 0, but there is no tptp file. How can i provide a RDF theory for the file? Or is this a bug?

Thanks for your significant work everyone.

tillmo:
The message `expected RDF theory for ...` is indeed a bug.

A workaround: it works if you explicitly select the logic translation (the `v2` option is just for logging the output file name): 
```
hets -o tptp -t CASL2TPTP_FOF -v2 example.casl
```

gibso:
Hello @tillmo,

thank you for your fast reply.

When i add the translation explicitly the minimal example is successfully converted into a tptp file.
But when i use more complex files, like the *[Examples/List.casl](https://github.com/spechub/Hets-lib/blob/master/Examples/List.casl)* File from the [Hets-lib](https://github.com/spechub/Hets-lib) repository, the bug (partly) occurs again.
I execute
```
hets -o tptp -t CASL2TPTP_FOF -v2 List.casl
```
and get the output:
```
Reading file file:///data/hets-lib/Examples/List.casl
Analyzing library Examples/List
Analyzing spec Elem
Analyzing spec List
Analyzing spec Monoid
Analyzing view MonoidAsList
Analyzing spec Nat
Analyzing spec NatList
Translated using comorphism CASL2TPTP_FOF : CASL -> TPTP
Writing file: ./List_Elem.tptp
### Warning:
for 'CASL2TPTP_FOF' expected sublogic 'SulFOL=E'
 but found sublogic 'eCPrenex=' with signature sublogic 'Atom'
Keeping untranslated theory
expected RDF theory for: ./List_List.tptp
Translated using comorphism CASL2TPTP_FOF : CASL -> TPTP
[...]

Issue #1706, opened 2017-05-24
axiom selection by sublogic


tillmo:
if the proof goal is formulated in some sublogic, try selecting those axioms that are in the sublogic, too. Works for example for some proofs in CASL. Currently, this principle is employed in CASL anyway, because the sort generation constraints are de-selected if the prover is a FOL prover. With axiom selection by sublogic, this ad-hoc stuff in CASL gets replaced by a systematic method.

PR #1907, opened 2019-07-21
1905 lossy translation

tillmo:
solves #1905 by providing a lossy flag `-Y` that makes translations to just leave out untranslatable axioms.
Note that #1706 must be solved before this one can be merged - otherwise, the proof support for CASL theories with sort generation constraints will be gone.

### DIFF
--- a/CASL/Sublogic.hs
+++ b/CASL/Sublogic.hs
@@ -34,6 +34,7 @@ module CASL.Sublogic
     , top
     , caslTop
     , cFol
+    , fol
     , cPrenex
     , sublogics_max
     , comp_list
@@ -181,6 +182,13 @@ cFol :: Lattice a => CASL_SL a
 cFol = caslTop
   { sub_features = NoSub -- no subsorting
   , has_part = False -- no partiality
+  }
+
+fol :: Lattice a => CASL_SL a
+fol = caslTop
+  { sub_features = NoSub -- no subsorting
+  , has_part = False -- no partiality
+  , cons_features = NoSortGen -- no sort generation constraints 
   }
 
 cPrenex :: Lattice a => CASL_SL a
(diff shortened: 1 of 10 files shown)


tillmo:
OK, I have managed to introduce some logic that avoids the problem with proving. However, still there is some problem when trying it out. Do you have an idea way?

mcodescu:
Could you please tell me what problem do you get, and perhaps also provide me the test file?

tillmo:
I just have found and fixed the problem. Could you please review again (this time you can give an official OK...), including trying some proofs for CASL theories?

mcodescu:
Now I get `ax_ax3` as the name of `ax3` in my test, or 'ax_ga_assoc___*__' for List.casl, and the behavior wrt. proving did not change. I think we can live with these names, but if we made somewhere (e.g. Ontohub or rest.hets.eu) the assumptions that comorphisms preserve names of sentences, we should be aware that this is no longer the case.
:+1:

tillmo:
I needed to add `ax_` because axiom names starting with a digit like `1_def_Nat` are a syntax error in TPTP. We could restrict these additions to those axiom names starting with a digit, at the risk that the mapping is no longer injective (`0` and `ax_0` will be both mapped to `ax_0`). What do you think?

mcodescu:
If a name `N` is not valid because it starts with a digit, we could check whether a sentence named `ax_N` is already part of the theory. If this is the case, we could generate a new name for the translation of `ax_N`.

tillmo:
yes, but in case that theories are united after translation (e.g. if a whole development graph is translated along a comorphism, which the old Hets GUI offers via the menus), there still could be clashes. So perhaps we should stay with the easy name transformation, which also has the advantage that its inverse can be computed easily.

mcodescu:
Fine with me, it's just names of sentences and we can still easily deduce their original name.

Issue #1882, opened 2019-05-06
IRIs display error


mcodescu:
for 
```
logic OWL
ontology DOLCE-Lite =
 <https://ontohub.org/spaceportal/LOA-cnr-it/DOLCE-Lite.owl>
end
```
with Show theory I get a theory that starts with

```logic OWL.NP-SHI

Prefix: : <http://www.loa-cnr.it/ontologies/DOLCE-Lite.owl#>
Prefix: owl: <http://www.w3.org/2002/07/owl#>
Prefix: rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
Prefix: rdfs: <http://www.w3.org/2000/01/rdf-schema#>
Prefix: xsd: <http://www.w3.org/2001/XMLSchema#>

Class: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#particular
Class: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quale
Class: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quality-space
Class: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#spatio-temporal-particular
AnnotationProperty: comment
EquivalentClasses: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quale,
                   //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#region
                   and //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#atomic-part-of some //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#region
EquivalentClasses: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quality-space,
                   //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#region
                   and //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#overlaps only (not //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quality-space)
EquivalentClasses: //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#spatio-temporal-particular,
                   //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#particular
                   and (//www.loa-cnr.it/ontologies/DOLCE-Lite.owl#endurant
                        or //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#perdurant
                        or //www.loa-cnr.it/ontologies/DOLCE-Lite.owl#quality)
[...]

Issue #1880, opened 2019-05-06
ontology with prefixes analyzed and displayed incorrectly


mcodescu:
With the ontology
```
logic OWL

ontology O1 = 
 Prefix:   : <http://test.org/O1#>
 Prefix:  a: <http://test.org/A#>
 Prefix:  b: <http://test.org/B#>
 
 Class: C
 Class: a:C
 Class: b:C
end
```
the result of `Show theory` is
```
logic OWL.ELQLRL-ALC

Prefix: : <<http://test.org/O1#>>
Prefix: a: <<http://test.org/A#>>
Prefix: b: <<http://test.org/B#>>

Class: C
```
Two issues: 
 - double < and > for prefixes
 - only one class C instead of C with different prefixes.

mcodescu:
For the issue with the double angles: the IRI of the value of the prefix has the flag `hasAngles` set to `true` (note that if the angles are missing we will correctly get a parsing error). Thus, when displayed, the method `iriToStringFull` adds the first pair of angles. The method `printPrefixes` in `OWL2/ManchesterPrint.hs` always prints angles, so thus we get the second pair. 

Unless we have an example when adding the second pair of angles is needed, I would not print them the second time.

mcodescu:
For the second issue: the static analysis tries to add `a:C` to the signature where `C` already exists. Checking equality is done via the case for full IRIs, when the `prefixName`s are not compared. Adding this comparison gets things to work. But I wonder if not rather `hasFullIRI` should be corrected, also because when showing the resulting signature the IRIs are again identified as full and for all three of them we get `C` instead of `C, a:C, b:C`, as the `prefixName` is not displayed for full IRIs.

tillmo:
Your solution for the first issue seems plausible to me.
Concerning the second issue, indeed the three different names should be displayed differently. I do not understand why the `prefixName` is not displayed for full IRIs. I think either the full IRI has to be displayed, or (if it is decomposed into prefix and local path) then the CURIE (= prefix + local path) generating the full IRI.

mcodescu:
My question was: is `a:C` a full IRI (with `isAbbrev` set to `True`) or an IRI that is not full? 

If it is a full IRI, we must make two changes: comparison between two IRIs should take into account the prefix and displaying of full IRIs should take into account the case when `isAbbrev` holds. 

If it is not a full IRI (my intuition), then `hasFullIri` should no longer identify `a:C` as a full IRI.

mcodescu:
[...]

Issue #1858, opened 2018-06-21
OWL prefixes are lost


tillmo:
When analysing
```
logic OWL
ontology O =
Class: C
DataProperty: dateOf
              Domain: C
              Range: xsd:date
```
and then outputting the resulting theory (e.g. with `-o th`), the result has a `Range: date`, which leads to an error, because `date` is an unknown datatype.

mcodescu:
We should test this again after the branch for iris is merged.

mcodescu:
Solved on #1909

PR #1909, opened 2019-07-31
Iris

mcodescu:
fixes for #1880, #1882.


mcodescu:
I can't reproduce the error of the test on my machine:
```
./hets --quiet -a none  ~/repos/Hets-lib/Ontology/Examples/OpenAALOntology.owl 
mcodescu@dell:~/repos/Hets2$ 
```
and calling Hets with `-g` gives the expected result.

mcodescu:
The IRI `http://www.w3.org/2001/XMLSchema#string` is not stored in the expected way: 
```
scheme:http:
no authority
path://www.w3.org/2001/XMLSchema#string
query:
fragment:
prefix:
isAbbrev:False
```
The prefix should be `http`, not empty, and the scheme should be empty. I'll try to fix this.

mcodescu:
In `OWL2/AS.hs/checkPredefAux` the `prefixName` of an IRI is compared with `http`. Now, I have changed this to `iriScheme` and compared with `http:` (First issue, the colon should not be saved here, or the assumption that it is saved should be made everywhere in the code). With this change, `http://www.w3.org/2001/XMLSchema#string` is recognized as a datatype. But `xsd:boolean` is no longer recognized. The method has to take both cases into account. As a hack in the last commit, I used the `iriScheme` if it's non-empty, otherwise the `prefixName` is used. With this hack, things work. Another bug is revealed: `xsd:boolean` is displayed `xsdboolean`. It looks to me that `Common/IRI.hs/iriToStringAbbrev` should add separators instead of assuming they are stored.

mcodescu:
#1858 also works as expected on this branch

tillmo (approved):
looks good and solves the issues, but please wait for travis (I have merged in master).

mcodescu:
Travis failes but this seems not to be related to the changes.

tillmo:
It seems that travis just has a timeout. I try again with #1943

tillmo:
The travis problem is this:
```
[163 of 839] Compiling OWL2.Parse       ( OWL2/Parse.hs, OWL2/Parse.o )
OWL2/Parse.hs:24:1: warning: [-Wunused-imports]
    The import of ‘Common.Id’ is redundant
      except perhaps to import instances from ‘Common.Id’
    To import instances alone, use: import Common.Id()
OWL2/Parse.hs:29:1: warning: [-Wunused-imports]
    The import of ‘sortId’ from module ‘Common.Token’ is redundant
Current RSS of 11379: 5.4G
Ooops, that's too much.
[164 of 839] Compiling OWL2.ManchesterParser ( OWL2/ManchesterParser.hs, OWL2/ManchesterParser.o )
make[1]: *** [hets_server-opt] Error 1
Makefile:510: recipe for target 'hets_server-opt' failed
make: *** [hets_server.bin] Terminated
Makefile:907: recipe for target 'hets_server.bin' failed
```
Actually, this failure does not make the test fail, because hets_server-opt is built successfully afterwards. However, this second attempt causes a longer running time that then causes the failure. The question is how to avoid this first failure.

tillmo:
The crucial lines are:
```
Current RSS of 11379: 5.4G
Ooops, that's too much.
```
But why does the memory consumption increase so much?

mcodescu:
After a restart, I don't get a timeout anymore. This is strange.

PR #1911, opened 2019-08-09
parser for GDOL, first attempt

mcodescu:
Not intended for merging, ignore any style issue or use of tracing messages.
Issues:
```logic OWL

pattern ValSet [Class: Val;  {Individual: v0} :: vS; ? ObjectProperty: greater] 
= %% all individuals vi from v0::vS become members of Val
let pattern OrderStep [Individual: vi; {Individual: vj} :: vS] =
     Individual: vj Types: Val Facts: greater vi
    then OrderStep[vj; vS]
    end
in   Individual: v0 Types: Val
then SimpleOrder[greater; Val] and OrderStep[v0; vS]
then { DifferentIndividuals: vs, vS
       Class: Val EquivalentTo: {vj, vS} } 
```
gives a parse error because the list is not the last element in the list of parameters. Moving the optional argument before it makes the example go through. The fact that the parameter following the list is optional has no relevance (removing the ? still gives an error).
I would have the same problem in instantiations `P[A; X::Xs; B]` so I have not implemented this yet.
```
logic OWL 

pattern GradedRels 
  [Class: S; Class: T; 
   ObjectProperty: p Domain: S Range: T; 
   Class: Val;
   {Individual: v Types: Val} :: valS  ]
= %% a sheaf of graded relations p[vi], one for each vi in v::valS
let pattern Step[Individual: x :: xs] =
     ObjectProperty: p[x] Domain: S Range: T SubPropertyOf: p 
     then Step[xs]
    end
    pattern Test [Class: C] =
     Class: C SubClassOf: S
    end
in Onto and Step[v::valS] and 
   { ObjectProperty: has[Val] Domain: T Range: Val } and Test[S]
```
goes through, but removing the optional end after the local patterns makes the parser fail (`skip` as a separator most likely not the best idea). At the level of `libItem`s, that also have an optional end, these are parsed with a recursive function that also tries to parse annotation. Should follow the same approach here?
```
logic OWL

pattern P[Class: C; ?ObjectProperty: p; Class: D] =
 Class: C SubClassOf: p only D
end
ontology O = P[A; ;B] 
```
How should I parse the missing optional parameter?
```
logic OWL
pattern SimpleOrder [Class: C] =
 TransitiveRelation[greater[C]; C] 
```
Using `hetIRI` to parse `greater[C]` does not work. I have used `compoundIriCurie`. But I wonder if `hetIRI` does what it should.

### DIFF
--- a/Common/Lexer.hs
+++ b/Common/Lexer.hs
@@ -291,6 +291,10 @@ commaT = asSeparator ","
 semiT :: CharParser st Token
 semiT = pToken $ string ";" << notFollowedBy (char ';')
 
+-- a double colon
+doubleColonT :: CharParser st Token
+doubleColonT = pToken $ string "::" << notFollowedBy (char ':')
+
 oBraceT :: CharParser st Token
 oBraceT = asSeparator "{"
 
--- a/Driver/WriteFn.hs
+++ b/Driver/WriteFn.hs
@@ -302,7 +302,7 @@ writeTheory ins nam opts filePrefix ga
                   (printTheory ms OWL2 $ OWL2.prepareBasicTheory th2) "\n"
             showDiags opts ds
             when (null sy)
-                $ case parse (OWL2.basicSpec Map.empty >> eof) f owltext of
+                $ case parse ((OWL2.basicSpec True) Map.empty >> eof) f owltext of
               Left err -> putIfVerbose opts 0 $ show err
               _ -> putIfVerbose opts 3 $ "reparsed: " ++ f
             writeVerbFile opts f owltext
--- a/Logic/Logic.hs
+++ b/Logic/Logic.hs
@@ -257,6 +257,9 @@ class (Language lid, PrintTypeConv basic_spec, GetRange basic_spec,
             Just p -> makeDefault (p, pretty)
          -- | parser for basic specifications
          parse_basic_spec :: lid -> Maybe (PrefixMap -> AParser st basic_spec)
+         -- | parser for macros
+         parse_macro :: lid -> Maybe (PrefixMap -> AParser st basic_spec)
+         parse_macro _ = Nothing
          -- | parser for a single symbol returned as list
          parseSingleSymbItem :: lid -> Maybe (AParser st symb_items)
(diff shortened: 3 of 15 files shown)


mcodescu:
Adding "pattern" to `Common/Keywords.hs/startingKeywords` and ";" and "::" to `Common/Keywords.hs/terminatingKeywords` makes first two issues work. This is not a solution because ";" is used in CASL as a separator. I wonder if it wouldn't have been easier to consider `sort s; op c : s` as two arguments instead of one (works with sequential semantics).

mcodescu:
With 
```
logic CASL

pattern FixPoint[sort s; op c : s; op f : s -> s] =
 . f(c) = c
end
```
I get only one parameter:
```
[OntoParam False (Annoted {item = Basic_spec Basic_spec [Annoted {item = Sig_items (Sort_items NonEmptySorts [Annoted {item = Sort_decl [s] nullRange, opt_pos = nullRange, l_annos = [], r_annos = []}] nullRange), opt_pos = nullRange, l_annos = [], r_annos = []},Annoted {item = Sig_items (Op_items [Annoted {item = Op_decl [c] (Op_type Total [] s nullRange) [] nullRange, opt_pos = nullRange, l_annos = [], r_annos = []}] nullRange), opt_pos = nullRange, l_annos = [], r_annos = []},Annoted {item = Sig_items (Op_items [Annoted {item = Op_decl [f] (Op_type Total [s] s nullRange) [] nullRange, opt_pos = nullRange, l_annos = [], r_annos = []}] nullRange), opt_pos = nullRange, l_annos = [], r_annos = []}] nullRange, opt_pos = nullRange, l_annos = [], r_annos = []})
```
This surprises me a bit, I thought ";" as a terminating would make me have 3 parameters. We can get 3 parameters using the notation of the paper `pattern FixPoint[{sort s}; {op c : s}; {op f : s -> s}] =...`, which is good news.

tillmo:
I think the reason is that the basic spec parser uses other terminating keywords than the structured spec and library parsers.

PR #1915, opened 2019-11-06
1914 allow the user to post files to the res tful interface

### LINKED ISSUE #1914: Allow the user to POST files to the RESTful interface
Instead of a coded IRI, the user should be allowed to POST a file that is then analysed by Hets.

Look at `parseRESTful` in `PGIP.Server`. In line 542, the coded IRI is bound to `libIri`, which is then passed to `getResponseAux` as part of a `NewDGQuery`. Instead of the coded IRI, there could be a file contents in the POST parameters (variable `requestBodyBS`). This should be stored into a local file, and its file path then passed as `libIri`. In order to obtain an IRI from the file path, prepend `file://` to the full path.

DerProfessor:
This shall fix #1914 .


Common/Counter.hs
```diff
@@ -0,0 +1,15 @@
+module Common.Counter where
```
tillmo: please remove this module, because it is not needed

PR #1923, opened 2020-02-04
1921 cache lib env

### LINKED ISSUE #1921: cache LibEnv
In `Driver.AnaLib` or `Static.AnalysisLibrary`, set up a cache for the LibEnv. 
* find out whether `Driver.AnaLib` or `Static.AnalysisLibrary` is the better place, possibly unify both (e.g. by using `ResultT` instead of `Maybe` at both places). I think that `PGIP.Server` should use `Driver.AnaLib`, and not, as currently done, `Static.AnalysisLibrary`. Our goal is that `PGIP.Server` will benefit from the cache.
* set up a cache for the `LibEnv`, use it only for those `LibName`s that the user wants to analyse and only if a `useCache` flag is set

DerProfessor:
This shall fix #1921 .

Because a cache already exists in the Hets codebase, this PR rather adds documentation for the existing cache and expands the file name in the cache. Now the absolute path is used when a local file on the system is analysed. This prevents multiple entires in the cache because prior to these changes the IRI in a rest call is directly used.

### DIFF
--- a/PGIP/Shared.hs
+++ b/PGIP/Shared.hs
@@ -1,11 +1,11 @@
 {-# LANGUAGE CPP, DoAndIfThenElse #-}
 {- |
 Module      : ./PGIP/Shared.hs
-Description : Provides various Hets resources.
+Description : Provides resources for caching requests to the RESTful interface.
 License     : GPLv2 or higher, see LICENSE.txt
 
 This module provides various Hets resources that are used t.e. for caching of an analysis
-library, proving data types etc.
+library during a request to the RESTful interface, proving data types etc.
 -}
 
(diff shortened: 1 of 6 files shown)


PGIP/Shared.hs
```diff
@@ -1,4 +1,12 @@
 {-# LANGUAGE CPP, DoAndIfThenElse #-}
+{- |
+Module      : ./PGIP/Shared.hs
+Description : Provides various Hets resources.
```
tillmo: this description is too unspecific. Better: "Provides resources for caching requests to the RESTful interface"

PR #1934, opened 2020-07-18
added hets-lib to package list

tillmo:
this is a PR for testing whether travis still works


jelmd (commented):
Actually hets-desktop as well as hets-server packages depend on the hets-libs package, so hets-libs should get pulled in/installed automagically when hets-{server|desktop} gets installed.  No need to install it explicitly.

tillmo:
OK, then this can be closed.

Issue #1941, opened 2020-09-22
make docs/index.html fails


tillmo:
```
make docs/index.html

Check log.haddock for results ...
haddock -o docs -h -s ../%F  \
            -t 'Hets - the Heterogeneous Tool Set' \
            -p Hets-Haddock-Prologue.txt --optghc=-DTAR_PACKAGE --optghc=-DNO_WGET --optghc=-DUNIX --optghc=-DUNI_PACKAGE --optghc=-DHASKELINE --optghc=-DHEXPAT --optghc=-DSERVER --optghc=-DWARP3 --optghc=-DHAXML --optghc=-DRDFLOGIC --optghc=-DCASLEXTENSIONS --optghc=-DGTKGLADE --optghc=-Wall --optghc=-fwarn-tabs --optghc=-fwarn-unrecognised-pragmas --optghc=-fno-warn-orphans --optghc=-fno-warn-unused-do-bind \
             OMDoc.hs COL.hs CspCASLProver.hs ExtModal.hs Proofs.hs CASL.hs Propositional.hs CMDL.hs Isabelle.hs CoCASL.hs s.hs TopHybrid.hs Hybrid.hs hets.hs CspCASL.hs GUI.hs Comorphisms.hs Logic.hs MMT.hs HasCASL.hs Modal.hs OWL2.hs Syntax.hs Taxonomy.hs Driver.hs ATC.hs Common.hs RelationalScheme.hs SoftFOL.hs Framework.hs Haskell.hs Static.hs CASL_DL.hs \
	>log.haddock 2>&1
Makefile:557: recipe for target 'docs/index.html' failed
make: *** [docs/index.html] Error 1
```

cmaeder:
Hi Till, I don't know why you assigned this issue to me. I had a short look into the file `log.haddock` and I saw the error 

``` 
ATC/Id.hs:28:1: error:
    Could not find module ‘ATerm.Lib’
```

However, I was not able to install the package `aterm` (using cabal) due to

```
src/ATerm/AbstractSyntax.hs:59:36: error:
    Not in scope: type constructor or class ‘TypeRep’
```

Building the binaries using stack and creating docs using haddock do no longer fit together.
Cheers Christian

cmaeder:
```
Persistence/Schema/OMSOrigin.hs:4:1: error:
    Failed to load interface for ‘Database.Persist.TH’
```

With a stack installed haddock I get the error above.

tillmo:
[...]

PR #1942, opened 2020-09-30
1941 make haddock work

tillmo:
shall fix #1941. Not ready for review yet

PR #1939, opened 2020-09-21
Make compilable under newer glib versions

TpmKranz:
This is what I had to change to build Hets on my Fedora 32 system.
The main problem seems to have been https://github.com/gtk2hs/gtk2hs/issues/276

Todos:
- update `resolver` to get rid of the new `extra-deps` that simply shadow older versions of existing packages
  - This is complicated by the fact that the `uni-*` packages break with newer `base` versions.
- remove dependency on (lib)glade(-dev)
- add a line for installing dependencies on Fedora

PR #1943, opened 2020-10-02
dummy commit

tillmo:
when compiling #1941 rebased on master, I get
```
[661 of 868] Compiling GUI.GtkUtils     ( GUI/GtkUtils.hs, GUI/GtkUtils.o )

GUI/GtkUtils.hs:549:3: error:
    • No instance for (TreeModelClass (Maybe (ListStore a)))
        arising from a use of ‘treeViewSetModel’
    • In a stmt of a 'do' block: treeViewSetModel view (Just store)
      In the expression:
        do { store <- listStoreNew listData;
             treeViewSetModel view (Just store);
             treeViewSetHeadersVisible view False;
             ren <- cellRendererTextNew;
             .... }
      In an equation for ‘setListData’:
          setListData view getT listData
            = do { store <- listStoreNew listData;
                   treeViewSetModel view (Just store);
                   treeViewSetHeadersVisible view False;
                   .... }
```
This PR is for checking whether this may be caused by commits on the master branch.


TpmKranz:
The signature change for `treeViewSetModel` was introduced in https://github.com/gtk2hs/gtk2hs/commit/b1f3bbe9f366a841777911fefba3bcdd4bb76128, leading me to believe you have a `gtk` version < 0.15 in your environment. Did you `make restack` after pulling the most recent master? #1939 introduced `gtk-0.15.5` as one of the `extra-deps`.

tillmo:
you are right, that was an old error that I could indeed solve with `make resrack`. I mixed things up.
The problem I want to tackle is that travis fails here: https://travis-ci.com/github/spechub/Hets/jobs/394044573

tillmo:
OK, f4ac445efe27b1d7cccfdd275715d33b74a744be (the current master) works. Now trying 43ee7ac001f5b116d3701979931c330aff883aa5

tillmo:
43ee7ac works. Now trying 28e56b0b6

tillmo:
Indeed, 28e56b0 failed - but why? Only two make targets have been renamed. Do they appear in the travis scripts?

tillmo:
The error is in the 5th travis job:
```
*** Error:

OWL2Parser.jar not found, check your environment variable: HETS_OWL_TOOLS

for Hets usage, use: hets -h

The command "test/hets-lib-check.sh" exited with 2.
```
Why is this related to 28e56b0 ?

tillmo:
Now trying 0200519b9, which is the same as PR #1942, just to see whether the results are the same.

tillmo:
Aha, need a dummy commit 8a1e0c2290381d83a5683639559a26efc7192512 - otherwise travis is not run again.

tillmo:
aha, this one has passed. So 0200519 should also pass, because the difference is only in `README.md`. It is problematic that travis results are not reproducible.

tillmo:
was for testing if travis works - not needed anymore

PR #1946, opened 2020-10-06
1940 aterms

cmaeder:
avoid to drop aterms just because the sources no longer compile with newer ghcs. Version 0.1.0.2. is also uploaded to https://hackage.haskell.org/package/aterm .The current ghc also works with this change.


tillmo (approved):
many thanks!

PR #1947, opened 2020-10-06
Fix var

cmaeder:
I got the following error without this change for `make`

```
ghc-pkg: cannot find package gtk
/bin/bash: Zeile 0: [: =: Einstelliger (unärer) Operator erwartet.
```


cmaeder:
I did not want to include the older commits

PR #1953, opened 2020-10-09
ensure that symbol names are legal in CASL via translation

### LINKED ISSUE #1895: name of OWL symbols along OWL2CASL
In OWL an individual may start with a number, e.g. `0Insignificant`. This is however not allowed in CASL. The translation from OWL to CASL should take care that proper CASL names are generated in these cases, e.g. `gn_0Insignificant`.

mcodescu:
Should fix #1895 and also replace non-CASL characters. Please let me know if I should allow more characters to remain as they are and if the solution to replace them is OK or if you would like something else.

### DIFF
--- a/OWL2/OWL22CASL.hs
+++ b/OWL2/OWL22CASL.hs
@@ -118,7 +118,21 @@ uriToIdM = return . uriToCaslId
 
 -- | Extracts Id from URI
 uriToCaslId :: IRI -> Id
-uriToCaslId urI =  stringToId $ showIRI urI
+uriToCaslId urI = 
+ let urS = showIRI urI
+     urS' = concatMap 
+            (\c ->  if isAlpha c || 
+                       isDigit c || 
+                       c == '\'' ||
+                       c == '_' 
+                   then [c] 
+                   else "_u" ) urS
+ in case urS of
+      x : _ -> 
+         if isDigit x then genName urS' 
+         else stringToId urS'
+      _ -> error "translating empty IRI" 
+           -- should never happen
 
 {- let
   repl a = if isAlphaNum a then [a] else if a/=':' then "_" else ""


OWL2/OWL22CASL.hs
```diff
@@ -118,7 +118,21 @@ uriToIdM = return . uriToCaslId
...
+     urS' = concatMap 
+            (\c ->  if isAlpha c ||
+                       isDigit c || 
+                       c == '\'' ||
+                       c == '_'
+                   then [c]
+                   else "_u" ) urS
+ in case urS of
+      x : _ -> 
+         if isDigit x then genName urS'
```
tillmo, line 132: this should be done if `x` is not a letter, since CASL Ids must start with a letter (aha, or a dot, see DOT-WORDS in the CASL reference manual)

OWL2/OWL22CASL.hs
```diff
@@ -118,7 +118,21 @@ uriToIdM = return . uriToCaslId
 
 -- | Extracts Id from URI
 uriToCaslId :: IRI -> Id
-uriToCaslId urI =  stringToId $ showIRI urI
+uriToCaslId urI =
```
tillmo, line 121: I think this function is not specific to this comorphism, so it could be better moved e.g. to Common.Id.

tillmo (changes requested):
looks good, but see comments to the code

Common/IRI.hs
```diff
@@ -1091,3 +1092,21 @@ addSuffixToIRI s i = if not $ null $ iriQuery i
...
+     urS' = concatMap 
+            (\c ->  if isAlpha c ||
+                       isDigit c || 
+                       c == '\'' ||
+                       c == '_'
+                   then [c]
+                   else "_u" ) urS
+ in case urS of
+      x : _ -> 
+         if not (isAlpha x || x == '.') then genName urS'
```
tillmo, line 1109: actually, `|| x == '.'` does not make sense here. If the IRI starts with a dot, the dot is replaced with `_u`, and the resulting identifier starts with `_u`, which is not legal in CASL. So the inital dot needs to be kept, but only if a `WORDS` in the CASL sense follows. 
But maybe the whole dot thing is not relevant. So you also could just omit `|| x == '.'`.

tillmo (changes requested):

Common/IRI.hs
mcodescu: I think if we omit the test from the code and an IRI should start with a dot, we would get an illegal CASL identifier. So the test should stay. The scheme and the prefix parts of an IRI cannot start with a dot, so the only way to get a dot at the beginning of an IRI is if it's in the IRI path. This should give us a CASL identifier via translation.

tillmo: If we omit the test from the code and an IRI should start with a dot, the dot will be replaced with `_u` and the whole identifier will be prefixed with `gn_`. So `.id` will be translated to `gn__uid`, which is a legal CASL Id.

Alternatively, we could translate `.id` to itself, which also is a legal CASL Id. However, this would require some more code.
 
If we leave things as they are now, `.id` will be translated to `_uid`, which is not a legal CASL Id.

tillmo: try yourself with:
```
ghci Common/IRI.hs
uriToCaslId  nullIRI{iriPath = stringToId ".id"}
```

mcodescu: You are right. I removed the test.

tillmo (approved):

PR #1965, opened 2020-10-17
profiling via make variable

tillmo:
This should make profiling easier. Keep `Makefile` and `var.mk` as they are. For compiling Hets with profiling, use
```
make PROFILE=on restack
make PROFILE=on hets         # or hets_server
```
The first line installs dependencies, the second line compiles Hets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@
 #
 # If the following environment variables are set, their values get passed to
 # the corresponding tool as is: GHC_PKG_FLAGS (ghc-pkg), GHC_FLAGS (ghc).
+# For profiling, call make with PROFILE=on 
 
 include var.mk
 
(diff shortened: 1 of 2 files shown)


mcodescu:
Also update https://github.com/spechub/Hets/wiki/Debugging-and-Testing-Hets

tillmo:
I tried this out with a fresh installation, and it failed with a strange error:
```
[649 of 709] Compiling Persistence.Schema ( Persistence/Schema.hs, Persistence/Schema.p_o )

<no location info>: error:
    Persistence/Schema.hs:1:1: fatal:
    cannot find object file ‘./Persistence/Schema/Enums.dyn_o’
    while linking an interpreted expression
Makefile:507: recipe for target 'hets_server' failed
```

mcodescu:
I get the same after a `make realclean`.

mcodescu:
But after a normal `make restack & make hets`, I don't get an error anymore. I wonder if we can't live with this for now, chances are small that someone will want to do profiling at first compilation.

tillmo:
I can confirm this. OK, then let us do it in this way.

tillmo:
@mcodescu can you review this one, please

mcodescu (approved):
The code looks good. Please make sure to remove the outdated comment that I already marked - I actually thought you would do this before I approve.
:+1:

Issue #1935, opened 2020-09-03
Hets cannot parse large OWL files


tillmo:
The Chebi ontology needs more than 400GB RAM to be parsed and translated to FOL.

tillmo:
Here the command that I have used:
```
hets +RTS -M400G -p -RTS -t OWL22CASL:CASL2TPTP_FOF -o tptp -v2 chebi.owl
```
You can get `chebi.owl` e.g. [here](https://www.ebi.ac.uk/ols/ontologies/chebi), size is 541MB.
It would be good to try out mere parsing and analysis first, and then the translation separately.

mcodescu:
Hets (used to ) parse(d) succesfully larger ontologies, e.g. from Bioportal. Have you tried to do some profiling?

tillmo:
OK, just parsing and analysing takes about 7 minutes and 36GB of RAM.

mcodescu:
Protégé also has problems opening the file, at least on my laptop:
```
Loading ontology from file:/home/mcodescu/Downloads/chebi.owl 
Finished loading file:/home/mcodescu/Downloads/chebi.owl 
An error occurred whilst loading the ontology at file:/home/mcodescu/Downloads/chebi.owl. 
Cause: GC overhead limit exceeded 
```

fabianneuhaus:
@mcodescu That's a well-known problem with Protege. You need to increase the heap size. 

https://protegewiki.stanford.edu/wiki/ChangingTheHeapSize

mcodescu:
@fabianneuhaus: thank you, I have tried to follow a different instruction and that did not work.

@tillmo: it seems to me that my hunch was wrong and the parser does not have a bug caused by some special form that the sentences of the ontology have. The most time and space is spent in comparisons and unions. We encountered a similar problem when we wanted to align the ontologies from Bioportal.

tillmo:
Do you think that the comparisons and unions are needed for static analysis, or for translation, or both?
I just have realised that my report of a successful static analysis needing 36GB RAM and 7 minutes might have been premature. I did not generate any output, so due to lazy evaluation, it might be that Hets has not even built the whole signature.

mcodescu:
I can only guess for now (I believe both creating the OWL signature and its translation). I think it would be better instead to do profiling for just parsing and analysing the ontology, and then for translation (or with translation). Yes, you could also add an output step to make sure the signatures get evaluated (some tracing message should do).

mcodescu:
[...]

PR #1970, opened 2020-11-16
1935 owl22casl sentence translation

mcodescu:
Should improve #1935.

### DIFF
--- a/OWL2/OWL22CASL.hs
+++ b/OWL2/OWL22CASL.hs
@@ -281,18 +281,18 @@ mapTheory (owlSig, owlSens) = let
         {- dTypes = (emptySign ()) {sortRel = Rel.transClosure . Rel.fromSet
                     . Set.map (\ d -> (uriToCaslId d, dataS))
                     . Set.union predefIRIs $ OS.datatypes owlSig} -}
-    (cSens, nSig) <- foldM (\ (x, y) z -> do
-            (sen, sig) <- mapSentence y z
-            return (sen ++ x, cSig)) -- uniteCASLSign sig y)) 
-                     ([], cSig) owlSens
-    return (foldl1 uniteCASLSign [nSig, pSig],  -- , dTypes],
+    (cSens, nSigs) <- foldM (\ (x, y) z -> do
+            (sen, y') <- mapSentence cSig z
+            return (sen ++ x, y ++ y')) -- uniteCASLSign sig y)) 
+                     ([], []) owlSens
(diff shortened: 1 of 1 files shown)


OWL2/OWL22CASL.hs
```diff
@@ -414,37 +415,37 @@ mapComObjectPropsList cSig mol props a b = do
...
 mapFacet :: CASLSign -> Int -> (ConstrainingFacet, RestrictionValue)
-    -> Result (CASLFORMULA, CASLSign)
+    -> Result (CASLFORMULA, [CASLSign])
 mapFacet sig var (f, r) = do
     con <- mapLiteral r
     let (fp, cf) = mkFacetPred con f var
-    return (fp, uniteCASLSign sig $ (emptySign ())
-            {predMap = MapSet.fromList [(cf, [dataPred])]})
+    return (fp, [(emptySign ())
+            {predMap = MapSet.fromList [(cf, [dataPred])]}])
```
tillmo, line 449: shouldn't `sig` appear in the resulting signature list?
why s the empty sig needed in the list?

OWL2/OWL22CASL.hs
```diff
@@ -495,77 +496,77 @@ mapCard b cSig ct n prop d var = do
         exactLst' = mkImpl (conjunct $ oPropsE ++ fOut) $ disjunct dlstM
         senAux = conjunct [minLst, mkForall qVarsE exactLst']
         exactLst =  if null qVars then senAux else mkExist qVars senAux
-        ts = uniteL $ [cSig] ++ s ++ s' ++ s''
+        ts = concat $ s ++ s' ++ s''
```
tillmo, line 499: why can `csig` be omitted here (and below)?

tillmo (commented):
My tests show that everything works fine. However, I do not understand the code at some places.

OWL2/OWL22CASL.hs
mcodescu: no, `sig` should actually be omitted from all these functions. I will do this. 
I kept the old formatting and it's hard to read, it should be the signature that you get by inserting (cf, dataPred) in the predMap of the empty signature.

OWL2/OWL22CASL.hs
mcodescu: csig is not needed, it was passed around just so we can return a signature when no change happens, and to be extended with the new symbols. I will remove it throughout the code.

mcodescu: Also, note that csig is added at the union in mapTheory anyways, and you don't want to have it there more than once.

tillmo (approved):

PR #1971, opened 2020-11-17
1924 tptp cons checkers

mcodescu:
In the GUI, we first select a prover and only then a comorphism path. On the paths to SoftFOL I get an error - `(Logic SoftFOL expected, but TPTP found)`. Perhaps we should have different options, e.g. Darwin-TPTP and Darwin-SoftFOL? Otherwise the code should be ready for review.

### DIFF
--- a/TPTP/ProveHyper.hs
+++ b/TPTP/ProveHyper.hs
@@ -1,9 +1,9 @@
 {- |
 Module      :  ./TPTP/ProveHyper.hs
 Description :  Interface to the theorem prover e-krhyper in CASC-mode.
-Copyright   :  (c) Dominik Luecke, Uni Bremen 2010
+Copyright   :  (c) Otto-von-Guericke University of Magdeburg, 2020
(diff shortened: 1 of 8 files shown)


TPTP/ProveHyper.hs
```diff
@@ -0,0 +1,302 @@
...
+Maintainer  :  luecke@informatik.uni-bremen.de
+Stability   :  provisional
+Portability :  needs POSIX
+
+Check out
+http://www.uni-koblenz.de/~bpelzer/ekrhyper/
+for details. For the ease of maintenance we are using e-krhyper in
+its CASC-mode, aka tptp-input. It works for single input files and
+fof-style.
+-}
```
tillmo, line 15: also here, the header needs adpation

tillmo (approved):
approved modulo the comments.
I think we do not need Darwin-TPTP and Darwin-SoftFOL, because SoftFOL should vanish soon.

PR #1974, opened 2020-12-11
display ref trees individually

### LINKED ISSUE #974: Some aspects of refinement trees do not work as expected
Reported by **till** and assigned to **mcodescu**
 Migrated from http://trac.informatik.uni-bremen.de:8080/hets/ticket/974

---

When working with a refinement tree
- parts of the tree appear many times - this is irritating
- some nodes have a menu "Show dependency diagram" although they do not corrspond to an arch spec
- Perhaps having arch spec nodes in a different color would help
- the menu "Show dependency diagram" does not work for arch spec nodes that are not the root of the tree (hence this problem should be solved before the first one above)

mcodescu:
Should fix #974:
 - instead of showing all refinement trees, let the user select a refinement specification and display only that tree
 - nodes that have a dependency diagram are colored blue
 - changed the way these nodes are computed to only roots of architectural specifications that have diagram stored in the development graph (e.g. the dependency diagram of an anonymous arch spec that is the specification of a unit in a larger arch spec is not stored and therefore cannot be displayed).

An example can be found here: https://ontohub.org/sandbox/RefinementExamples.casl. I've just noticed there is a bug with composition of refinement pointers, so please use this file not the one in Hets-lib.

### DIFF
--- a/GUI/ShowRefTree.hs
+++ b/GUI/ShowRefTree.hs
@@ -34,9 +34,32 @@ import Static.PrintDevGraph
 import Static.History
 
 import qualified Data.Map as Map
+import qualified Data.Set as Set
+import Data.Char
+
+lookup' :: (Ord a) => Map.Map a b -> a -> b
+lookup' x y = Map.findWithDefault (error "lookup': key not found") y x
 
 showRefTree :: LibFunc
 showRefTree gInfo = do
+  ost <- readIORef $ intState gInfo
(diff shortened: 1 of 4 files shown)


tillmo (approved):
Looks good!
One comment: while the ability to choose a tree is certainly useful, I think it is still useful to be able to show the whole refinement tree, but without duplications. But maybe this is a separate issue.

PR #1992, opened 2021-05-05
1984. OWL2 AS printer

daniilsvc:
PrintOWL2AS.hs - printer
ASKeywords.hs - keywords for functional syntax


tillmo:
the issue should be only closed after the pull request has been reviewed and merged

daniilsvc:
There was one syntax mistake on my local branch in AS.hs
"AsyMetricObjectProperty" instead of "AsyMMetricObjectProperty" and it is unfortunately included in pull request. I think @mcodescu can reject this request and I recreate one.

tillmo:
you can just add a new commit. This will be integrated into this pull request. No need to create a new one.

OWL2/PrintOWL2AS.hs
```diff
@@ -0,0 +1,608 @@
...
+-- | print Root
+instance Pretty PrefixDeclaration where
+    pretty (PrefixDeclaration prName iri) =
+        keyword "Prefix"
+        <> sParens ((text prName) <> (text "=") <> pretty iri)
+
+instance Pretty Ontology where
+    pretty (Ontology mOnt mVerIri dImpDocs ontAnns axioms) =
+        keyword "Ontology"
+        <> sParens (hsep . concat $
```
mcodescu, line 597: Use `vsep` instead of `hsep`, otherwise you get the whole ontology on one very long line.

mcodescu (commented):
I have just a few comments so far, things seem good otherwise. This branch should be merged with the parsing for easier review - right now I've been building abstract syntax ontologies by hand. But let's wait on @tillmo 's comment from my review of the other branch

mcodescu:
I can parse (with Björn's code), print with this code and then open the resulting ontology in Protege (if it has no imports, see my comment below). This worked for several ontologies and in principle the two branches could be merged on 1945 after fixing the issues from my review.

mcodescu:
If we have an IRI that has abbreviation, we check that its `prefixName` appears in the `PrefixDeclaration` argument of the printing function. If it does, we display the prefixName and the `iFragment`, separated by `:`. If it does not, we display the IRI like it is currently displayed. For example, with the prefix declaration:
```
Prefix( a: = <http://example1.org/onto#> )
```
we display 
```
IRI
scheme: http
authority:IRIAuth {iriUserInfo = "", iriRegName = "example.org", iriPort = ""}
path:/onto
query:
fragment:hasFather
prefix: a
iFragment: hasFather
isAbbrev: True
```
as `a:hasFather`, but
```
IRI
scheme: http
authority:IRIAuth {iriUserInfo = "", iriRegName = "example.org", iriPort = ""}
path:/onto
query:
fragment:hasFather
prefix: b
iFragment: hasFather
isAbbrev: True
```
as `http://example.org/onto#hasFather`, because `b` is not a declared prefix.

OWL2/PrintOWL2AS.hs
```diff
@@ -0,0 +1,788 @@
...
+    keyword "Ontology"
+    <> sParens (vsep . concat $
+    [[ontNameDoc], [importedDocs], ontAnnsDocs, axiomsDocs])
+    where
+        ontAnnsDocs = map (printAnnotation pds) ontAnns
+        axiomsDocs = map (printAxiom pds) axioms
+        versionIriDoc = maybe empty (printIRI pds) mVerIri
+        ontNameDoc = maybe empty (\ontvalue -> hsep [printIRI pds ontvalue,
+            versionIriDoc]) mOnt
+        importedDocs = keyword "Import"
```
mcodescu, line 783: If there are no imports, don't display `Import(  )`.

mcodescu:
Given the ontology:
```
Prefix( a: = <http://example1.org/> )

Ontology(
  Declaration( ObjectProperty( a:hasFather ) )
  Declaration( ObjectProperty( <http://example2.org/onto1#hasMother> ) )
)
```
I parsed it with Björn's code then displayed it after deleting the prefixes. It still displayed `a:hasFather` instead of expanding the prefix. I think you should remove the prefix and the iFragment from the iri before pretty printing it.

mcodescu:
Also `isAbbrev` must be set to `False`.

OWL2/PrintOWL2AS.hs
daniilsvc: done

daniilsvc:
Now Printer/Parser passed your case and all *.ofn tests in OWL2/tests

daniilsvc:
It works only if in _IRI.hs_ we substitute lines 382 - 384 from
```
   n <- ncname
   c <- string ":"
   return $ n
```

to

```
n <- optionMaybe ncname
c <- string ":"
return $ maybe "" id n
```
which of course I didn't commit in this branch.

mcodescu:
With the file here: https://ontohub.org/sandbox/soma.ofn, I get:
```
Ontology( :

          Import( : : )

```
The ontology name is not displayed correctly, the imports are merged to only one import declaration and not displayed correctly. Moreover, in several places I get sentences of the form: `ObjectPropertyDomain( :meets : )`, or even `SubClassOf( :BodyMovement ObjectAllValuesFrom( : ObjectUnionOf( ObjectSomeValuesFrom( : : ) ObjectSomeValuesFrom( : ObjectIntersectionOf( : ObjectSomeValuesFrom( : : ) ) ) ) ) )`. I think this task should be done once the IRI parsing is finished.

OWL2/PrintOWL2AS.hs
```diff
@@ -0,0 +1,795 @@
...
+            versionIriDoc]) mOnt
+        importedDocs
+            | null dImpDocs = empty
+            | otherwise =
+                keyword "Import"
+                <> sParens(hsep $ map (printIRI pds) dImpDocs)
+
+instance Pretty OntologyDocument where
+    pretty (OntologyDocument prefDecls ont) = 
+        (hsep . map pretty $ prefDecls) $+$ printOnt prefDecls ont
```
mcodescu, line 795: separate the prefix declarations using `vsep` instead of `hsep`

OWL2/PrintOWL2AS.hs
```diff
@@ -0,0 +1,795 @@
...
+    where
+        ontAnnsDocs = map (printAnnotation pds) ontAnns
+        axiomsDocs = map (printAxiom pds) axioms
+        versionIriDoc = maybe empty (printIRI pds) mVerIri
+        ontNameDoc = maybe empty (\ontvalue -> hsep [printIRI pds ontvalue,
+            versionIriDoc]) mOnt
+        importedDocs
+            | null dImpDocs = empty
+            | otherwise =
+                keyword "Import"
```
mcodescu, line 790: why not display them as: 
```
Import (IRI1)
Import (IRI2)
...
Import(IRIn)
```

PR #1996, opened 2021-05-12
Extended IRI to store expanded IRI as well as CURIE

b-gehrke:


mcodescu:
Have you tried to compile Hets and got no error? Or do you plan to fix possible such compilation errors later?

b-gehrke:
> Have you tried to compile Hets and got no error? Or do you plan to fix possible such compilation errors later?

I actually made the changes on master and it compiled there (please feel free to verify). Then switched to 1945 and tested it with starting GHCI for OWL2/AS.hs and all modules could be loaded.

mcodescu (approved):
:+1:

PR #2000, opened 2021-07-09
Removed most 'try' usages

b-gehrke:
The implementation of the OWL AS Parser heavily relies on `try`s. All of those that could be spared got removed and instead of skipping spaces at the beginning of a parser all parsers now skip spaces at the end.


mcodescu (approved):
Looks good :+1:

Issue #1993, opened 2021-05-05
in OWL2 Logic instance, switch to functional abstract syntax


tillmo:
- in `OWL2.Logic_OWL2`, change instance declarations to the new functional abstract syntax
- comment out all comorphisms and provers (and maybe more?) in order to make Hets compile
-  restoring comorphisms and provers will be done in subsequent issues
This is part of #1945

mcodescu:
Update the following:
- static analysis 
- sublogic analysis
- XML and XMLConversion
- conservativity, extract module, taxonomy

Last two items can be commented out first.

PR #2004, opened 2021-07-28
1983 extend owl2 manchester parser to new as

### LINKED ISSUE #1983: extend the Parsec parser for Manchester syntax to the whole new abstract syntax
part #1945

b-gehrke:
Closes #1983 

Implemented parser for MS syntax to new AS.


b-gehrke:
The `Individuals` section cannot be yet parsed

b-gehrke:
Even with an updated version of the bioportal ontologies there are still some errors in some (converted) ontologies. E.g. BIM.owl:539 :

Annotation defined as integer in .owl file
```xml
<owl:DatatypeProperty rdf:about="&BIM2;pmcid">
    <rdfs:comment rdf:datatype="&xsd;integer">
        PubMed Central Identifier of a scientific docuemnt.
    </rdfs:comment>
</owl:DatatypeProperty>
```
in converted MS has no quotes and therefore produces invalid MS which cannot be parsed.
```owl2
DataProperty: <http://cbakerlab.unbsj.ca:8080/sebi/BIM.owl#pmcid>
    Annotations: 
        rdfs:comment PubMed Central Identifier of a scientific docuemnt.
```

All bioportal ontologies (that are converted to valid MS Syntax) can be parsed (Except of cause SWRL Rules)

OWL2/scripts/runManchester.hs
```diff
@@ -24,7 +24,7 @@ import Text.ParserCombinators.Parsec
 processFile :: String -> IO ()
 processFile file = do
   str <- readFile file
-  case runParser (basicSpec << eof) () file str of
+  case runParser (parseOntologyDocument << eof) () file str of
```
mcodescu, line 27: I get an error here, the PrefixMap is missing.

mcodescu (commented):
I have taken a look at the code and I've tested it on several ontologies. It seems to be working right, modulo a minor change. I missed the discussion on Wednesday, so I have to ask: have you sorted the issue about performance on larger ontologies?

Also, please add the usual comment at the beginning of `OWL2/ParseMS.hs`.

OWL2/scripts/runManchester.hs
b-gehrke: Fixed

b-gehrke:
> I have taken a look at the code and I've tested it on several ontologies. It seems to be working right, modulo a minor change. I missed the discussion on Wednesday, so I have to ask: have you sorted the issue about performance on larger ontologies?

We archived a speedup of up to 105. The main reason was one function used in the MS printer which was doing many comparisons on IRI's.

This parser can parse the largest bioportal ontology GEXO.omn with 273MB in about 3.5min.

mcodescu (approved):
:+1:

b-gehrke:
In the current state HETS won't compile because we haven't switched to the new AS in the Logic instances. As this is the next task (#1993) @daniilsvc and me are working on, I suggest overwriting the checks and doing manual merge once more.

PR #2018, opened 2021-09-28
Fixing iri formatting and owl2 static analysis

b-gehrke:
- Fixed `iriToStringAbbrev` (therefore also `instance Show IRI`) to show an abbreviated IRI only with `prefixName` and `iFragement` as those properties hold all information using the extended data type of #1996  .
- Adjusted OWL2 printers to print the empty prefix if abbreviated
- Adjusted check for predefined prefixes using the extended data type of #1996 
- Fixed OWL2.StaticAnalysis to report only the Object- or DataPropertyExpressions that are not declared when checking for a list


mcodescu (approved):
:+1:

PR #2024, opened 2021-10-14
Fixed using MS snippets in dol file

### LINKED ISSUE #2023: Allow MS Snippets in .dol file
Part of #1945 
With AS being the new default language, MS snippets in .dol files must be marked as such. Implement support for this.

b-gehrke:
Closes #2023


b-gehrke:
I used the example in the issue which worked fine for me. Using your example here - without `end` - works just as well. Now it should also work with a closing `end` which also allows the following snippet to be processed:
```
logic OWL serialization Manchester

ontology O = 
 Prefix: o: <https://example.com/>
 Class: o:A
 Class: o:C
    SubClassOf: o:A

ontology P =
 Class: D

end
```

mcodescu (approved):
:+1:

Issue #1955, opened 2020-10-10
Integerate Hermit as prover for OWL2


tillmo:
see http://www.hermit-reasoner.com/

mcodescu:
Checking consistency (on Ubuntu you might have to use full path of the jar file):
```
java -jar /home/mcodescu/Software/HermiT/HermiT.jar -l /home/mcodescu/Software/HermiT/consist.owl -k
Exception in thread "main" org.semanticweb.owlapi.reasoner.InconsistentOntologyException: Inconsistent ontology
	at org.semanticweb.HermiT.Reasoner.throwInconsistentOntologyExceptionIfNecessary(Unknown Source)
	at org.semanticweb.HermiT.Reasoner.checkPreConditions(Unknown Source)
	at org.semanticweb.HermiT.Reasoner.isSatisfiable(Unknown Source)
	at org.semanticweb.HermiT.cli.CommandLine$SatisfiabilityAction.run(Unknown Source)
	at org.semanticweb.HermiT.cli.CommandLine.main(Unknown Source)

```
Proving that an ontology is entailed by the other:
```
java -jar /home/mcodescu/Software/HermiT/HermiT.jar --premise=https://ontohub.org/sandbox/premise.owl --conclusion=https://ontohub.org/sandbox/conclusion.owl -E
```

tillmo:
good. Since Hermit can prove entailment, it should have priority over Konclude #1875 and Robot #1956

tillmo:
Here some test example, save it under test_owl.dol:
```
logic OWL

ontology O =
 Class: A
 Class: B SubClassOf: A
 Class: C SubClassOf: B
then %implies
 Class: C SubClassOf: A
end
```
and save this script as test_owl.hpf:
```
use test_owl.dol
dg-all auto 
dg basic O__E1 
prover Hermit
prove-all 
```
and then in a shell call
```
hets test_owl.hpf
```

erithacus-rubecula:
I tried to run the hets test_owl.hpf but I get the error: 'no applicable prover with name "Hermit" found'.  I thought it might be the missing HERMIT_PATH but writing it into etc/environment didn't help. What am I doing wrong here?

mcodescu:
For me it worked by removing `lib/` from `hermitJar,` as the jar file is directly in the folder where Hermit is.

erithacus-rubecula:
I removed the lib/ part when we talked about it last week, so that can't be the cause.

mcodescu:
[...]

PR #2026, opened 2021-10-30
1955 hermit proover integration

erithacus-rubecula:
requested pull for hermit integration branch.


tillmo:
@mcodescu can you please review this one?

mcodescu:
Yes.

mcodescu:
The last problem reported by me in the ticket #1955 is not fully solved: I see that you have added the case to `analyseOutput` but the other change was not made and therefore I get the same error when trying to disprove something. I will mark the change I had to make directly in the code. 

You should also remove the tracing messages.

OWL2/ProveHermit.hs
```diff
@@ -0,0 +1,281 @@
...
+           : (if doEntail then ["--premise=file://"++timeTmpFile, "--conclusion=file://"++entFile, "-E"] 
+              else ["-k", timeTmpFile]) ++ words opts
+      case entail of
+        Just c -> writeFile entFile c
+        Nothing -> return ()
+      mex <- timeoutCommand secs "java" args
+      trace (show mex ++ "\n" ++ show args) (removeFile timeTmpFile)
+      removeFile timeTmpEnt
+      when doEntail $ removeFile entFile
+      return $ fmap (\ (_, outS, errS) -> if outS == "true\n" then (True, outS, errS) else (False, outS, errS)) mex
```
mcodescu, line 199: Using False here make the tool report it cannot find the prover. It worked for me as expected with:
`return $ fmap (\ (_, outS, errS) -> (True, outS, errS)) mex`

mcodescu (changes requested):

mcodescu (approved):
:+1:

Issue #1936, opened 2020-09-03
output prover-independent proof (hyper)graphs


tillmo:
For provers that output proofs, parse them and represent proof steps as nodes and dependencies between them as edges or hpyeredges. The  data structure for proof graphs should be logic and prover independent, while the logic and prover specific parts come in as node and (hyper)edge labels (which could be e.g. strings, but they could also contain sentences for the logic).
Possibly hypergraphs and not graphs are needed, because then hyperedges can be annotated with proof rules (if available). On the other hand, since each node probably only has one and only one incoming hyperegde, one could also use normal graphs and use multiple simple edges instead of one hyperedge. The proof rule then needs to be attached to the node.

Issue #1502, opened 2015-06-10
integrate first-order induction prover into Hets


tillmo:
e.g. Inka

PR #2032, opened 2021-11-22
Introduce useful proof trees/proof graphs and Zipperposition as CASL prover

TpmKranz:
Addresses #1936 for EProver and SPASS.


TPTP/Prover/SPASS/ProofParser.hs
```diff
@@ -43,7 +46,32 @@ parseOutput namedGoal = foldl parseIt (Nothing, [], False, midnight)
...
+              (True, Right (clId, _, parents, _)) -> [(p, -clId-1, i) | (i,p) <- zip [0..] parents] ++ (-clId-1,clId,-1):clEdges
+              _ -> clEdges
+          ) where parsedProofLine = runParser proofLine () "" line
+
+
+proofLine :: Parser (Int, String, [Int], String)
+proofLine = do
+ clauseId <- decimal
+ char '['
+ decimal -- What's this?
```
tillmo, line 63: this is the split level, see "OUTPUT" at https://webspass.spass-prover.org/help/options.html

tillmo: here more about the notion of split level: https://core.ac.uk/download/pdf/205222916.pdf

TpmKranz: I guess I'd reverse-engineered the output before I found the documentation you linked and didn't bother looking at the OUTPUT section after I finally found it in the context of TPTP output. Thanks for pointing this out!

It seems to me that the labelling and therefore the split level is just bookkeeping to enable non-chronological backtracking. As far as I can tell (which, admittedly, is not very far), the proof graph makes it apparent at which levels the splits happened, since only the ones that contributed to finding the refutation are recorded in the output proof. I'd therefore conclude that we don't need to include this information in a proof graph.

What do you think?

TpmKranz:
Now also addresses #1502

I built the ZP integration on top of the proof graph branch, since ZP can output TPTP proofs as well, which was almost immediately useful in debugging the integration.

tillmo (changes requested):
looks good. It seems that there are some filles copied from elsewhere. Could you please give attribution to the author(s)?

tillmo:
@TpmKranz could you please provide the changes?

TpmKranz:
> @TpmKranz could you please provide the changes?

Thanks for the reminder! I hope that now everything is attributed/licensed properly.

tillmo (approved):

PR #2037, opened 2021-12-16
Changed default serialization for OWL back to Manchester Syntax.

b-gehrke:
Changed default serialization for OWL back to Manchester Syntax.
Manchester Syntax is used in published papers and will be generally used more often than Functional Syntax.


mcodescu (approved):
:+1:

Issue #1949, opened 2020-10-07
double name for omn file


mcodescu:
With
```
Prefix: : <http://www.ontohub.org/godp/>
Ontology: <Role1>
Class: Role
```
the generated node is named "Role1Role1"

mcodescu:
`getOntologyIRI` in `OWL2/XML.hs` causes this. We should check whether it's a matter of the base not being set correctly or if the mentioned function should be corrected.

PR #2038, opened 2022-01-05
1949 update java owlapi

daniilsvc:
This branch updates java OWL API  version to 4.5.20. It solves issue #1949 - now if ontology IRI is relative it is made absolute by adding prefix "urn:absolute" during parsing into xml.

I also tried it with 5th version of OWL API (5.1.19), but it keeps ontology IRI relative. So I suggest we update it to version 4.5.20 for now.


mcodescu:
I see you are still making changes, please let me know when this is ready for review.

daniilsvc:
We have decided last meeting, that this branch will update OWL API to 5th version despite the fact that it doesn't solve #1949. So it is done.

Currently it works fine on my machine with java-1.8.0-openjdk.

mcodescu:
Works for me. Please try to fix the travis test failure before we can merge.

daniilsvc:
I think we are free to merge it, @mcodescu

mcodescu (approved):
:+1:

daniilsvc:
I've fixed a bug that didn't allow to parse anonymous ontology, e.g. TMA.owl. This PR is ready for review.

OWL2/java/build.xml
```diff
@@ -5,7 +5,13 @@
     <property name="builddir"   value="build"/>
     <property name="classesdir" value="${builddir}/classes"/>
     <property name="target-vm"  value="1.7"/>
-    <property name="owlapi-jars" value="lib/owlapi-osgidistribution-3.5.2.jar lib/guava-18.0.jar lib/trove4j-3.0.3.jar"/>
+    <property name="owlapi-jars" value="lib/owlapi-distribution-5.1.19.jar
+      lib/slf4j-api-1.7.25.jar lib/javax.inject-1.jar lib/caffeine-2.9.2.jar
+      lib/commons-io-2.11.0.jar lib/hppcrt-0.7.5.jar lib/slf4j-log4j12-1.7.25.jar
+      lib/log4j-1.2-api-2.12.1.jar lib/log4j-api-2.12.1.jar lib/log4j-core-2.12.1.jar
```
jelmd, line 11: https://logging.apache.org/log4j/2.x/ checked ???

daniilsvc: That is not our case. This vulnerability is about building flexible reports that use data from some shared server via LDAP. For example we can use expression "${jndi:ldap:<pattern>}" and it will substitutes <pattern> with data fetched from remote server. <pattern> can be some code for execution - and this is potential for code injection.

Our configuration is very simple, we use trivial pattern layout: <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n" /> that just format time/date and print it into console. We never try to access remote server so we are free from these bug.

Anyway, I've looked through dependencies and updated log4j2 to the recommended 2.17.1 version.

jelmd: @daniilsvc: Thanx!

Wrt. to the vulnerability: Well, not really - just try out this tiny thing with the simple log4j.xml of this PR:
```
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

/*
 * javac -cp log4j-api-2.12.1.jar:log4j-core-2.12.1.jar:. Demo.java
 * java -cp log4j-api-2.12.1.jar:log4j-core-2.12.1.jar:. -Dlog4j2.configurationFile=log4j.xml Demo '${jndi:ldap:github.com/foo}'
 */
class Demo {
    private static Logger logger = LogManager.getLogger();

    public static void main(String[] args) {
        if (args.length == 0) {
            System.err.println("java ... inputStr");
            return;
        }
        logger.info("This is strange: {}", args[0]);    
    }
}
``` 
You'll probably get an `WARN Error looking up JNDI resource ....` error. The 2nd thing is: Even if it would really depend on the logger, only: a config file is used and this can be changed at any time.

Anyway, the real question to answer wrt. this is: How is logging used. If one obeys best coding practice and never logs "received" input 1:1, etc., but escapes it as needed, there wouldn't be any harm wrt. to this vulnerability, even when using the old libs. But if e.g. a web app logs the unescaped username for failed logins, voilà. So good SW vendors look through there own code and check, how logging gets used. But analyzing all the 3rd part libs they use, is a big burden and thus many prefer the easy way and say, just upgrade and you are probably safe.

BTW: Wondering, how the list of new jars got determined. Can't find any logger use in **/*.java. If I compile owlapi from source, there is a hugh list of dependencies aka *.jar. - certainly not all (e.g. testing stuff) are not needed. However, I couldn't find a README, or dependency tree, or something like that,  which recommends, which one has to take over for his own apps? Can you shed a light on it, plz? Last but not least: AFAICS owlapi seems to use slf4j, only. So I wonder, where the dependency for log4j gets introduced? Is this somewhere documented?

daniilsvc: 1) You are right - we don't need log4j.

Simple slf4j is enough. That was silly mistake of mine. Since version of 1.6 we can use no-operation logger `slf4j-nop.jar`. Accidently on my disc I had file `slf4j-nop-1.7.35.jar` but in `build.xml` I used `lib/slf4j-nop-1.7.25.jar`. Because of these typo I simply ignored the `jar` on the disc and repeatedly got pesky error:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
every time I run `java -jar OWL2/OWL2Parser.jar -o xml test.xml test.omn`. Which is just a warning and output xml is generated successfuly, but Hets fails in such cases anyway. So I stopped any attempt to fix it and decided to proceed with log4j2 binding `slf4j-log4j2.jar`.

My mistake. We don't need log4j2 indeed.

2) As I understand, we need this big list of dependencies.

If I leave only next dependencies in `build.xml`: `owlapi-distribution-5.1.19.jar`, `slf4j-api-1.7.25.jar`, `slf4j-nop-1.7.25.jar`, `guava-31.0.1-jre.jar`, `trove4j-3.0.3.jar` - I am not able to run parser due to the next error (I skip stack trace):
```
Exception in thread "main" java.lang.NoClassDefFoundError: javax/inject/Inject
```
That is because class `org.semanticweb.owlapi.apibinding.OWLManager` has next import:
```java
import org.semanticweb.owlapi.utilities.Injector;
```
and this _Injector_ comes from `javax.inject-1.jar`. So we need it.

And this is true for other _jars_ in the list. This is actually how I got them first time - I simply tried to correct all `NoClassDefFoundError`s.

If you look into `pom.xml` of `owlapi-distribution-5.1.19` you will see all these dependencies listed in `<dependencies>`. And version 3 of owlapi doesn't have those dependencies - that's why our previous `build.xml` was more compact.

Of course using Maven it is enough just to mention `owlapi-distribution-5.1.19` as a dependency and other libraries will be fetched from central repository into `~/.m2/repository/` automatically. But we use ant in Hets so we need to collect all _jars_ manually.

Btw, you can run `mvn dependency:tree` to check which libraries `owlapi-distribution` depends on.

Tbh I've just realized, that I also need to extend this list of dependencies in `build.xml`. Because currently it contains only those that I found 'experimentally'  by fixing java exceptions. Of course I'm not going to include testing libraries (`mockito-core`, `findbugs`, `junit-jupiter`, `hamcrest-core`) and any library that is already included by `maven-shade-plugin`. It seems like the easiest way would be to run `mvn dependency:copy-dependencies` at some project with `owlapi-distribution-5.1.19` dependency and copy everything from `target/dependency` to `OWL2/java/lib`.

I'll do these updates shortly. 

### To sum up
Next commit (comming soon) will do next things:
1. Remove `slf4j-log4j2` binding and proceed with no-operation logger.
2. Extend dependency list to meet all dependencies and not only those, that I found via exceptions trying to run `java -jar OWL2/OWL2Parser.jar -o xml test.xml test.omn`.

jelmd: @daniilsvc Thanx a lot for for your insights.

1) logging
Wrt. to slf4j: Not sure, whether it is a good thing to use the noop implementation. Log messages are usually important wrt.  troubleshooting/forensic analyzes. So if no sophisticated logging and no debug messages  are needed, the simple-log implementation (slf4j-simple-*.jar) might be an option. For minimal dependencies/overhead using the JDK log implementation aka JUL (slf4j-jdk14-*.jar) - the 1.4 just says, it is available since JDK1.4 - would be a good alternative, too. If a little bit more sophisticated logging and bling bling is needed, the logback implementation should be used. The log4j implementation is IMHO the worst choice - the companies I worked for and myself kicked this rusty thing with all its backward compat. baggage  almost 20 years ago for several reasons, in favor of its successor: slf4j + logback ;-)  Last but not least: log4j is officially EOLed, again.

    Anyway, wrt. Hets I think JUL would be the sufficient.

2) dependencies
Yepp, I'm afraid that we need to include more. Actually I hoped, that you have found an article or a nice tool which allows one to reduce the list to a minimum. Unfortunately the pom stuff in the owlapi source tree doesn't group the deps into build, test and runtime requirements - so more or less useless because a catch all or nothing op :( .

   I'll try to find/write something, which checks the imports and perhaps we get at least rid off the test and build requirements.

daniilsvc: As I said, I've just run `mvn dependency:copy-dependencies` on some project that has `owlapi-distribution-5.1.19` in dependencies and copy everything from `target/dependency` into Hets `OWL2/java/lib`. It also get transitive dependencies like `httpcore` and `httpclient`. I've checked it with dependency list [here](https://search.maven.org/artifact/net.sourceforge.owlapi/owlapi-distribution/5.1.19/bundle) and seems like they match (except of testing libs).

jelmd: OK. That's one way. However, just because others use XY, it doesn't automatically imply, that their deps satisfy the requirements of a different project, too. E.g. they might have excluded or replaced some parts because they don't need it, there are collisions with the project itself, etc.. I think the best is to go back to the source and to try to find out, how it gets build etc. - this way one can be more or less sure, that the application doesn't explode on some corner case.

So what seems to work wrt. owlapi and doesn't require a lot of work/fine tuning is something like this:
```
rm -rf /tmp/lib{,.osgi}
cd /var/tmp
git clone https://github.com/owlcs/owlapi
cd owlapi/distribution
mvn --settings=${HOME}/.config/mvn/settings.xml -DoutputDirectory=/tmp/lib -DincludeScope=runtime dependency:copy-dependencies
cd ../osgidistribution
mvn --settings=${HOME}/.config/mvn/settings.xml -DoutputDirectory=/tmp/lib.osgi -DincludeScope=runtime dependency:copy-dependencies
```

A `diff -bur /tmp/lib /tmp/lib.osgi` shows, that both maven configs are not in sync wrt. lib versions. Since most projects use semantic versioning, it should be easy to decide, whether replacing the version in `lib/` by the one in `lib.osgi/`. Usually they are backward compatible, therefore the latest version should be preferred. So according to the described way my diff to your selection would be:
```
-caffeine-2.8.1.jar
+caffeine-2.8.6.jar
-httpcore-nio-4.4.12.jar
-httpcore-osgi-4.4.12.jar
+httpcore-nio-4.4.5.jar
+httpcore-osgi-4.4.5.jar
-jcl-over-slf4j-1.7.22.jar
+jcl-over-slf4j-1.7.30.jar
-owlapi-distribution-5.1.19.jar
+owlapi-api-5.1.19.jar
+owlapi-apibinding-5.1.19.jar
+owlapi-compatibility-5.1.19.jar
+owlapi-impl-5.1.19.jar
+owlapi-oboformat-5.1.19.jar
+owlapi-parsers-5.1.19.jar
+owlapi-rio-5.1.19.jar
+owlapi-tools-5.1.19.jar
-slf4j-api-1.7.5.jar
+slf4j-api-1.7.30.jar
```

I'm not sure wrt. `checker-qual-3.8.0.jar` and `error_prone_annotations-2.3.4.jar` - I guess this stuff is for static analysis, only and could be dropped. Not sure, where you got `httpcore-*4.4.12.jar` but IMHO ok to use it.

BTW: Are there any valid xml files, which can be used for testing?

daniilsvc: `httpcore-osgi-4.4.12.jar` comes as dep for [jsonld-java](https://search.maven.org/artifact/com.github.jsonld-java/jsonld-java/0.13.0/bundle), which comes as dep for [rdf4j-rio-jsonld](https://search.maven.org/artifact/org.eclipse.rdf4j/rdf4j-rio-jsonld/3.7.4/jar).

`httpcore-nio` comes as dep for [httpcore-osgi](https://search.maven.org/artifact/org.apache.httpcomponents/httpcore-osgi/4.4.12/bundle) itself.
```
BTW: Are there any valid xml files, which can be used for testing?
```

Do you mean ontology to test Hets with? Actually we have a lot of them already downloaded directly from bioportal on our local computers but they remain untracked. You could try smth like `./hets https://ontohub.org/bioportal/TMA.owl` that will make Hets to download the ontology (it's a small one). Or write a simple ontology that meets some particular requirement (relative ontology IRI, for example) that we use in issue mentioned above into file `test.omn`:
```
Prefix: : <http://www.ontohub.org/godp/>
Ontology: Role1
Class: Role
``` 
and run either `./hets test.omn` or java parser directly `java -jar OWL2/OWL2Parser.jar -o xml output.xml test.omn`

Btw currently some tests failed and I guess this is because of logger. It outputs some warning in stdout/stderr but Hets treats it as error. This is actually more Hets issue than dependency issue.

daniilsvc: Thank you, @jelmd!

Right now Hets works with java parser output next way:
https://github.com/spechub/Hets/blob/852a28b7328be808ccddc14345642d16529a6c99/OWL2/ParseOWL.hs#L47-L53

So everything printed in stderr treated as error. I actually thought to change this pattern matching into smth like:
```haskell
(ExitSuccess, msg) -> do
   print "OWL2Parser stderr:"
   print msg
...
_ -> fail $ "process stop!"
```
At least we need to do smth like that, because otherwise we never remove tmp file produced by Parser in case of error appearance. But I think this problem should be reflected in separate issue. After all Hets has been working like that for years before. So, I suppose proper log config is the best solution that can be done in this PR.

jelmd: Well, for troubleshooting the temp files are a valuable source.  Allowed me to run java ... directly. But yes, if verbosity level is 0, they should be automagically removed ;-). However, I agree, should be done in another PR (as well as indenting for java related stuff: tabs only).

IMHO ok to squash merge.

tillmo: > BTW: Previously I tried `OWL2/test/run.sh` which all fail. @tillmo: Are these tests of any use anymore? If not, cleanup to avoid confusion?

I think we should repair and keep them.

Issue #2039, opened 2022-02-09
improve performance of OWL static analysis


mcodescu:
On https://github.com/spechub/Hets/commits/1935_profiling_large_OWL_files, I have used strict hash maps instead of plain maps and I got a performance increase before finding a bug in OWL22CASL and abandoning the enterprise. 

We could try:
 - strict maps
 - strict hash maps
 - strict/strict hash sets
 - strict folds (foldl') -- definite candidate
 - perhaps strict field records
 
See https://making.pusher.com/top-tips-and-tools-for-optimising-haskell/

Issue #1636, opened 2016-05-12
OWL ontologies are read/written unnecessarily often


tillmo:
When Hets parses an OWL file, it does this twice, which is a waste of resources.
See https://github.com/spechub/Hets/files/260688/test-results.txt

tillmo:
Hets could pass a file IRI of the tmp file to the OWL API.

PR #2040, opened 2022-02-09
1636 hets performance issues

daniilsvc:
This branch is responsible for solving issues #1636 and #2039.
So far is done: no more redundant downloads of ontology from remote server if uri is provided to Hets.
TODO: revise static analysis. It seems that it makes sense to turn some Hets modules into strict mode / use strict data structures.


daniilsvc:
@mcodescu, I think this branch is ready for review. Here we:
1. prevent Hets from downloading Ontology multiple times
2. replace current Hets xml-parser (Expat and parser in Text.XML.Light.Input) with [Xeno xml-parser](url)

We still have gradually increasing memory consumption in case of big ontology. It happens in two places, which we assume work as we expect:
1. ontology document is created from parsed xml-file (function xmlBasicSpec in OWL2.XML)
2. static analysis in OWL2.StaticAnalysis creates intermediate entities number of which proportional to number of IRIs in ontology.

mcodescu (approved):
:+1: Looks good to me.

PR #2042, opened 2022-02-16
Expanded Manchester Syntax Datatype keywords

### LINKED ISSUE #1457: owl/omn keyword resolution
Looking at my commit message https://github.com/spechub/Hets/commit/437db42248d4e92184bf34417043a8fe891c751b I wonder, why "integer" was expanded to "http://www.dfki.de/sks/hets/ontology/unamed#integer" (now "http://www.hets.eu/ontology/unamed#integer").

The idea was to expand the keyword "integer" to " http://www.w3.org/2001/XMLSchema#integer" without the need to write "xsd:integer", but that is probably not in the spirit of OMN (OWL Manchester Notation).

So currently "integer" (without prefix) is our own keyword that is different from "xsd:integer" since only the expanded part is used when comparing IRIs. Setting the prefix for "http://www.hets.eu/ontology/unamed#integer" to "xsd:" makes the IRI inconsistent, although this is hardly noticable, since only the expanded part is compared and only the local part "integer" is (ever?) printed out.

b-gehrke:
Closes #1457


mcodescu (approved):
:+1:

Issue #2044, opened 2022-03-04
test scripts: on error exit with code != 0


jelmd:
The following test scripts exit always with an `exit code == 0`. However, they should do this on success, only (i.e. otherwise exit with `0 < exit_code < 256`).
- OWL2/scripts/runTest.hs
- Comorphisms/test/sublogicGraph.hs

This is required to make tests reliable and not depend on parsing the output ...

PR #2045, opened 2022-03-05
make check: fail on any error

jelmd:
Scripts used for checking now propagate errors so that `make check` now fails as well.

Per default `make check` (and related scripts) stop immediately on any error for easier troubleshooting. To get all errors in a row (i.e. do not stop after an error) run `FAIL_EARLY=0 make check`.

TBD/See also: #2044

### DIFF
--- a/OWL2/tests/test10.dol
+++ b/OWL2/tests/test10.dol
@@ -1,4 +1,4 @@
-logic OWL serialization Functional serialization Functional
+logic OWL serialization Functional
 
 spec a = 
 Prefix(:=<http://example.com/#>)
--- a/OWL2/tests/test10.dol
+++ b/OWL2/tests/test10.dol
@@ -1,4 +1,4 @@
-logic OWL
+logic OWL serialization Functional serialization Functional
 
 spec a = 
 Prefix(:=<http://example.com/#>)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,6 +179,7 @@ jobs:
           echo 'man-db man-db/auto-update boolean false' | \
               sudo debconf-set-selections
           sudo rm -f /var/lib/man-db/auto-update
+          sudo apt-get update
           sudo apt-get install ksh93 || sudo apt-get install ksh
           sudo apt-get install ${{ env.MISSING_GUI_PKGS }}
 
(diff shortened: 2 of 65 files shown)


jelmd:
The problem is now fixed. However, before merging it into the main branch the tests should be fixed for obvious reasons ;-)

b-gehrke:
I'd like to work on the failing tests but as you are merging from your forked repository I cannot push to this branch. Can you please push your branch to the Hets repository?

jelmd:
> ... I cannot push to this branch. 

Hmmm, strange. As a member of this org it should be possible ....

> Can you please push your branch to the Hets repository?

Yepp, no poblem: done.

b-gehrke:
> Hmmm, strange. As a member of this org it should be possible ....

It worked now. Probably some git remote stuff that I messed up.

Most OWL2 tests success now. There are 2 failing. The are OWL2/tests/{4.dol,5.dol}. The problem appears to be such a renaming:
```
with Class Woman |-> Female, NamedIndividual Peter |-> John
```

The current datatype for SymbMap Items. Allows only to specify one EntityType followed by multiple IRI pairs
```haskell
data SymbMapItems = SymbMapItems ExtEntityType [(IRI, Maybe IRI)]
```

Therefore, the code above is parsed as follows
```haskell
SymbMapItems (EntityType Class) [(A,Just B),(NamedIndividual,Nothing)]
```
@tillmo, @mcodescu It seems like `Syntax.Parse_AS_Structured.parseMapOrHide` should do a second call to `OWL2.symbItemsMap`    (Might be related to  64bbf05?). Maybe we restrict the symbol IRIs not to be a curie with omitted empty prefix and an entity type keyword? For example, if one wants to rename a class `:NamedIndividual`, one would write `with Class :NamedIndividual |-> OtherIndividual`. Writing `with Class NamedIndividual |-> OtherIndividual` would then be forbidden.

tillmo:
yes, the problem is that we currently do not allow the `EntityType` be specified for each symbol pair. This should be changed. See function `symbMapItems` in `CASL.SymbolParser`.

OWL2/tests/test10.dol
```diff
@@ -1,4 +1,4 @@
-logic OWL
+logic OWL serialization Functional serialization Functional
```
mcodescu, line 1: why twice here?

mcodescu (changes requested):
Looks good, minor change needed.

OWL2/tests/test10.dol
b-gehrke: Removed

mcodescu (approved):
:+1:

PR #2046, opened 2022-03-16
Changed return code of test scripts to be not 0 on failure.

### LINKED ISSUE #2044: test scripts: on error exit with code != 0

b-gehrke:
Closes #2044 

Haskell test scripts always returned 0 - even on failed test cases. Now they return `ExitFail`.


mcodescu (approved):
:+1:

b-gehrke:
@jelmd It seems like the checks are not running anymore. Do you have an idea what could be the reason?

jelmd:
Hmmm, perhaps GHA limits. Wondering what you see on https://github.com/settings/billing ?

jelmd:
@b-gehrke Seems to work again. Used "rebase on master" on the WebUI to trigger a rebuild. So your local copy of this branch should be dropped and if needed checked out again.

If this happens again, just wait some time, add|remove a `#` in the Makefile and commit/push to trigger the CI run.

b-gehrke:
> Hmmm, perhaps GHA limits. Wondering what you see on https://github.com/settings/billing ?

It says I have used 0 out of the included 2000 min for GHA.

>@b-gehrke Seems to work again.

Thank you!

>If this happens again, just wait some time, add|remove a # in the Makefile and commit/push to trigger the CI run.

Good to know! What do you think is a reasonable amount time to wait?

jelmd:
> > Hmmm, perhaps GHA limits. Wondering what you see on https://github.com/settings/billing ?
> 
> It says I have used 0 out of the included 2000 min for GHA.

OK. Found out, that there is one for the [organization](https://github.com/organizations/spechub/settings/billing) as well, but this shows also `0`and the generated report for the last 30 days is empty. So I guess it was not a limit but general github problem.
 
> > If this happens again, just wait some time, add|remove a # in the Makefile and commit/push to trigger the CI run.
 
> Good to know! What do you think is a reasonable amount time to wait?

Good question! I would probably check heise.de ;-) and if there are no related news, after 10 min. If it still doesn't work, after an hour and than double the interval ...

PR #2051, opened 2022-03-30
2047 change to ghc 8.6.5

tillmo:


tillmo:
@daniilsvc please take into account that Ubuntu 22.04 comes with ghc 8.8 (and also Ubuntu 20.04 has "ghc (8.8.1+dfsg1+is+8.6.5+dfsg1-3)", whatever that means. So if you switch to ghc 9, please test whether this is also compatible with ghc 8.8.

daniilsvc:
@tillmo, seems like now hets can be successfully compiled. You're free to check it.

tillmo:
we will wait for #2062, then test, and then this PR could be merged.

tillmo (commented):
since #2062 and #2066 have been closed now, this one can be merged now.

daniilsvc (approved):

PR #2053, opened 2022-04-07
Allowed OWL2 keywords as CURIE if prefixed with colon

### LINKED ISSUE #2052: OWL2 Manchester Syntax parser wrongly rejects CURIES with keywords
In OWL2 Manchester Syntax CURIES with empty prefix can be written without the leading colon if the fragment is not a reserved keyword. Therefore, a CURIE with empty prefix and keyword as fragment *must* be written with a leading colon.

The OWL2 MS parser currently does not check if a colon is present and rejects all CURIEs with empty prefix and keyword as fragment.

For example:
This is not allowed
```
Datatype: string
```
and should be written as

```
Datatype: :string
````

Currently, both are rejected by the parser

b-gehrke:
Closes #2052

### DIFF
--- a/OWL2/Parse.hs
+++ b/OWL2/Parse.hs
@@ -34,6 +34,7 @@ import Text.ParserCombinators.Parsec
 import Control.Monad (liftM2)
 import Data.Char
 import qualified Data.Map as Map
+import Data.Maybe (isJust)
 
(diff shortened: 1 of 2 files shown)


mcodescu (approved):
:+1:

PR #2058, opened 2022-04-28
1865 recognition of owl sublanguages

b-gehrke:
Recognition of OWL2 DL.

### DIFF
--- a/Logic/Logic.hs
+++ b/Logic/Logic.hs
@@ -853,15 +853,12 @@ class (StaticAnalysis lid
          -- no logic should throw an error here
          addOmdocToTheory _ _ t _ = return t
 
+         -- | sublogic of a theory
+         sublogicOfTheo :: lid -> (sign, [sentence]) -> sublogics
+         sublogicOfTheo _ (sig, axs) = 
+            foldl lub (minSublogic sig) $ map minSublogic axs
 
--- | sublogic of a theory
-sublogicOfTheo :: (Logic lid sublogics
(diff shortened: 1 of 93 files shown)


b-gehrke:
@mcodescu Please review and check the specified sublogics in the test files, too.

One test is currently failing which is fixed in #2045

mcodescu (commented):
The code looks really good. Is the test where I've commented the one that is fixed in 2045?

b-gehrke:
> The code looks really good. Is the test where I've commented the one that is fixed in 2045?

The test failing is `10/test3.omn` which is due to an error in parsing SWRLRules in the ontology.

I don't know why `11/test5.ofn` is not working properly in the GUI, but I'll investigate.

b-gehrke:
Apparently, the UI calls `Static.GTheory.sublogicOfTh` to retrieve the sublogic. More importantly, choosing an appropriate prover also relies on this function.
This function does not consider the ontology as a whole but rather folds over the sublogic of each __individual__ sentence. Therefore, global OWL2 restrictions cannot be discovered.

mcodescu:
I think this needs to be changed. @tillmo, do you agree?

b-gehrke:
Changing this would require a change in sublogic calculation/implementation in all language, wouldn't it?

mcodescu:
Yes, it would.

mcodescu (approved):
:+1: Yes, this is what I expected to see.

PR #2061, opened 2022-06-23
downgrade Java OWLAPI to version 4

daniilsvc:
Downgrade Java OWLAPI to version 4


mcodescu (commented):
The changes seem fine but you should see why you get `java.lang.IllegalStateException: Optional.get() cannot be called on an absent value` when running the tests.

mcodescu (approved):
:+1:

PR #2068, opened 2022-07-13
2056 fix factpp

### LINKED ISSUE #2056: use FaCT++ binary
The fact++ Ubuntu package contains `/usr/bin/FaCT++` and this should be used instead of `OWLFact.jar`.

b-gehrke:
Closes #2056 

By default, the empty prefix was not added. This lead to it missing in the `.owl` file handed over to the Java component which uses the OWL API to parse it. With no empty prefix defined but used, no axioms could be parsed by the OWL API.

### DIFF
--- a/ATC/Grothendieck.der.hs
+++ b/ATC/Grothendieck.der.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE UndecidableInstances, FlexibleInstances, OverlappingInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE FlexibleInstances #-}
 {- |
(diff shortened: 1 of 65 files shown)


mcodescu (approved):
:+1:

b-gehrke:
The introduction of the empty prefix at the level of logic graphs lead to more problems than expected. I reverted the commit and moved it to a seperate issue #2087. For the owl API and Fact++ to work with empty prefixes I added a simple workaround in OWL2.XMLConversion.
@mcodescu do you agree with these changes?

mcodescu (approved):
Yes we could do with this workaround :+1:

PR #2067, opened 2022-07-10
make hets compatible with ghc 8.8.4 and lts-16.31

daniilsvc:
This code compatible with both ghc 8.6.5 and ghc 8.8.4 (if we compile without stack), and also with both lts-14.27 and lts-16.31 (if we compile with stack). Related to #2064.

Issue #2065, opened 2022-06-30
build Focal Ubuntu packages for Hets


tillmo:
wait for #2062, #2066 and then merging of #2051 first

tillmo:
now, all prerequisites are there, so you can work on this one.

jelmd:
Is it possible to get rid off all the warnings? It is pretty hard to search in a file with 4100+ lines for an error. E.g. the 180 `Defined but not used` and 116 `* is redundant` warnings.

Another question: this requires GTK3 now? GTK2 + GLADE is not supported anymore? If so, there are some errors, which need to be fixed as well - see 2nd attached file.

tillmo:
We stick with GTK2. The switch to GTK3 should be a separate issue, see #1966

jelmd:
OK, then we need https://hackage.haskell.org/package/glade . Unfortunately this cannot be build - seems to be outdated wrt. cabal:

```
SetupWrapper.hs:84:49: error:
    * Couldn't match type `Distribution.System.Platform -> String'
                     with `[Char]'
      Expected type: String
        Actual type: Distribution.System.Platform -> String
    * Probable cause: `exeExtension' is applied to too few arguments
      In the second argument of `(<.>)', namely `exeExtension'
      In the second argument of `(</>)', namely
        `"setup" <.> exeExtension'
      In the expression: setupDir </> "setup" <.> exeExtension
   |
84 |     setupProgFile    = setupDir </> "setup" <.> exeExtension
   |                                                 ^^^^^^^^^^^^

SetupWrapper.hs:87:24: error:
    * Data constructor not in scope: Version :: [Integer] -> [a0] -> t
    * Perhaps you meant one of these:
        variable `mkVersion' (imported from Distribution.Version),
        variable `version0' (imported from Distribution.Version),
        variable `noVersion' (imported from Distribution.Version)
   |
87 |     useCabalVersion  = Version [1,8] []
   |                        ^^^^^^^

SetupWrapper.hs:113:34: error:
    * Data constructor not in scope:
        PackageName :: [Char] -> PackageName
    * Perhaps you meant one of these:
        variable `packageName' (imported from Distribution.Package),
        variable `mkPackageName' (imported from Distribution.Package),
        variable `unPackageName' (imported from Distribution.Package)
    |
113 |       let cabalDep = Dependency (PackageName "Cabal")
    |                                  ^^^^^^^^^^^

SetupWrapper.hs:117:23: error:
[...]

PR #2069, opened 2022-07-21
remove majority of warnings

daniilsvc:
Remove majority of warnings appearing during Hets compilation. Warnings about deprecated imports of `network` package will be fixed in #2067. Motivated by https://github.com/spechub/Hets/issues/2065#issuecomment-1178183691. Works both with and without stack.

`Non-exhaustive pattern match` warning in file `Static/PrintDevGraph.hs` was handled by stub `_ -> Doc.empty`. Can be fixed later.

### DIFF
--- a/ATC/Grothendieck.der.hs
+++ b/ATC/Grothendieck.der.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE UndecidableInstances, FlexibleInstances, OverlappingInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE FlexibleInstances #-}
 {- |
 Module      :  ./ATC/Grothendieck.der.hs
 Description :  manually created ShATermConvertible instances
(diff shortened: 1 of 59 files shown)


mcodescu (commented):
Looks good. Could this be fixed too?
```

TopHybrid/AS_TopHybrid.hs:2:5: warning:
    -XOverlappingInstances is deprecated: instead use per-instance pragmas OVERLAPPING/OVERLAPPABLE/OVERLAPS
  |
2 |   , OverlappingInstances #-}
  |     ^^^^^^^^^^^^^^^^^^^^

ATC/Grothendieck.hs:1:55: warning:
    -XOverlappingInstances is deprecated: instead use per-instance pragmas OVERLAPPING/OVERLAPPABLE/OVERLAPS
  |
1 | {-# LANGUAGE UndecidableInstances, FlexibleInstances, OverlappingInstances #-}
```

daniilsvc:
These two files are in `.gitignore`

mcodescu (approved):
Right. OK to merge then. :+1:

daniilsvc:
Is it OK to change *.der.hs files? They are not in gitignore and I found out that actual .hs files can be changed during compilation.

mcodescu:
Yes, the .der.hs generate the .hs files, it is right to change them.

PR #2077, opened 2022-07-24
Changed nesy pattern syntax

### LINKED ISSUE #2075: Change of NeSy pattern syntax
Currently, we can write `class`, `class[id]` or `id`. Please switch to `class`, `id : class`, because this conforms better with OWL2 ABox syntax. (The third option with `id` should go.)

b-gehrke:
Closes #2075

### DIFF
--- a/NeSyPatterns/AS.der.hs
+++ b/NeSyPatterns/AS.der.hs
@@ -32,10 +32,10 @@ import Data.Data
 -- DrIFT command
 {-! global: GetRange !-}
 
--- | nodes are of form: ontology_term[id]
--- | both components are optional, but at least one must be present
+-- | nodes are of form: id : ontology_term
+-- | id is optional
 data Node = Node {
-    ontologyTerm :: (Maybe Id.Token),
+    ontologyTerm :: Id.Token,
     nesyId :: (Maybe Id.Token),
     nodeRange :: Id.Range
   } deriving (Show, Typeable, Data, Eq, Ord)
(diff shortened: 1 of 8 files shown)


NeSyPatterns/AS.der.hs
```diff
@@ -32,10 +32,10 @@ import Data.Data
 -- DrIFT command
 {-! global: GetRange !-}
 
--- | nodes are of form: ontology_term[id]
--- | both components are optional, but at least one must be present
+-- | nodes are of form: id : ontology_term
+-- | id is optional
 data Node = Node {
-    ontologyTerm :: (Maybe Id.Token),
+    ontologyTerm :: Id.Token,
```
mcodescu, line 38: Why not IRI instead of Token?

b-gehrke: I remember @tillmo mentioning this should be done in a later task. Is this still true or can it done in this?

mcodescu: I have a workaround in the colimit computation, so I guess it could stay. @tillmo what do you think?

tillmo: If you have a workaround for colimits, let us stay with Token here, and have a separate issue for changing it into IRI.

mcodescu (approved):
Looks good to me.

tillmo (changes requested):
looks good, but I would like to be able to write `m :  Model`. Currently, I need to write `m: Model`, i.e. without spaces.

b-gehrke:
> looks good, but I would like to be able to write `m : Model`. Currently, I need to write `m:Model`, i.e. without spaces.

Fixed

PR #2078, opened 2022-07-26
isInfixOf import

daniilsvc:
Accidentally remove import of `isInfixOf` function. Preprocessor remove one of `#ifdef` branches that's why I got warning `function not used` but it is needed indeed.


mcodescu (approved):
:+1:

PR #2088, opened 2022-08-20
2057 use pellet binary

### LINKED ISSUE #2057: use Pellet binary
Also see #2056. The pellet Ubuntu package contains /usr/bin/pellet and this should be used instead of lib/pellet-cli.jar.

b-gehrke:
Changed from pellet java file to use pellet binary shipped with hets packages.

Note: first three commits are cherry-picked from #2068 

Closes #2057.


mcodescu (approved):
It's been a while since I was able to use Pellet :+1:

PR #2090, opened 2022-08-21
2087 add empty prefix to logicgraph

### LINKED ISSUE #2087: Empty prefix does not exist in logicgraph
In `Static.AnalysisLibrary.hs` the function `defPrefixGlobalAnnos` adds a default empty prefix. However, this is only used in analysis. The prefix map expected from Parsers for example do not see this empty prefix. Therefore, the empty prefix has to be defined explicitly.

Example: The parser in `OWL2.ParseMS.parseOntologyDocument`  will get an empty prefix map with the following .dol file. It is expected to contain the empty prefix.
```
logic OWL serialization Manchester
spec O =
    Class Person
end
```

b-gehrke:
Closes #2087

### DIFF
--- a/Comorphisms/LogicGraph.hs
+++ b/Comorphisms/LogicGraph.hs
@@ -38,6 +38,7 @@ Assembles all the logics and comorphisms into a graph.
 
 module Comorphisms.LogicGraph
     ( logicGraph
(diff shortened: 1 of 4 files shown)


mcodescu (approved):
:+1:

PR #2097, opened 2022-09-12
fix errors while compiling dtdtohaskell

daniilsvc:
Problems compiling Hets on master. Next sequence of commands fails:

1. `make distclean`
2. `make utils/DtdToHaskell` or `make Isabelle/IsaExport.hs`

This PR intends to fix it.


mcodescu (approved):
Looks OK :+1:

PR #2102, opened 2022-09-14
remove superfluous checks

tillmo:
The comorphism builds all the infrastructure for translating impredicative higher-order sentences, but then forbids just these. I have just removed the superfluous checks.


mcodescu (approved):
This was strange. :+1:

PR #2129, opened 2022-12-15
1790 api

b-gehrke:
haskell hets API with first steps for Python API


b-gehrke:
Remove old Changes and see if still works

Hets.hs
```diff
@@ -0,0 +1,31 @@
+module Hets (
```
b-gehrke, line 1: Rename to HetsAPI

Hets/Commands.hs
```diff
@@ -0,0 +1,178 @@
...
+libFlatHiding = globalCommandR Hiding
+
+libFlatHeterogen :: LibName -> LibEnv -> Result LibEnv
+libFlatHeterogen = globalCommandR Heterogeneity
+
+qualifyLibEnv :: LibName -> LibEnv -> Result LibEnv
+qualifyLibEnv = globalCommandR QualifyNames
+
+
+loadLibrary :: FilePath -> HetcatsOpts -> IO (Result (LibName, LibEnv))
```
b-gehrke, line 143: Unify with read and analyse

Hets/Python.hs
```diff
@@ -0,0 +1,126 @@
+module Hets.Python (
```
b-gehrke, line 1: For G_theory show sign as set as strings

b-gehrke: use `Logic.Logic.sym_of` and show the symbols

Docker/hyphen.Dockerfile
```diff
@@ -0,0 +1,14 @@
+
+FROM ubuntu:22.04
```
b-gehrke, line 2: Change to :lts

Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
```
tillmo, line 1: what purpose does this file have? Please add a comment, or remove the file

Driver/ReadFn.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
```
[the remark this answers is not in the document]
b-gehrke: It's not. Removed it.

Setup.hs
b-gehrke: Comment added. This file is only used for library building. Cabal generates this file by default when creating a new library project.

tillmo (approved):

PR #2136, opened 2023-02-01
2131 for the python api have a uniform datatype for sentences

### LINKED ISSUE #2131: for the Python API, have a uniform datatype for sentences
Since existential types cannot be accessed in Python, implement a tree-like datatype for sentences, such that sentences can be mapped into this datatype.

b-gehrke:
Closes #2131

Adds rules for deriving `GHC.Generic`, `Data.Aeson`s `FromJSON`, `ToJSON`, `FromJSONKey`, and `ToJSONKey` instances for all datatypes that are also convertible to ATerms.

With these changes, JSON can be used as interchange format between the python api and the haskell api.

### DIFF
--- a/ATC/Sml_cats.hs
+++ b/ATC/Sml_cats.hs
@@ -1189,7 +1189,7 @@ instance ATermConvertibleSML SPEC where
         case aterm of
             (ShAAppl "basic" [ aa ] _) ->
                 let
-                aa' = from_sml_ShATerm (getATermByIndex1 aa att)
+                aa' = from_sml_ShATerm (getATermByIndex1 aa att) :: CASLBasicSpec
                 aa'' = G_basic_spec CASL aa'
                 in group (Syntax.AS_Structured.Basic_spec aa''
                           nullRange) group_flag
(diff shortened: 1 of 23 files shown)


tillmo (changes requested):
somehow, I would expect that the change from
```
deriving (Show, Eq, Ord, Typeable, Data)
```
to 
```
deriving (Show, Eq, Ord, Typeable, Data, Generic)
```
should also be made for the abstract syntaxes of all logics. Why is this not needed here?

b-gehrke:
This deriving is added with the new DrIFT-rule `Json`. For each ADT `A` it adds a derivation for `Generic` and default instances for `ToJSON` and `FromJSON`:

```hs
deriving instance GHC.Generics.Generic A
instance Data.Aeson.ToJSON A where
instance Data.Aeson.FromJSON A where
```

b-gehrke:
@tillmo Now the `ToJSON` and `FromJSON` classes are required for everything that require the `PrintTypeConv` class in `Logic.Logic`. This includes sentences. The derivings for Abstract Syntax elements are added with a new DrIFT Rule.

tillmo (approved):

PR #2138, opened 2023-02-22
2108 python api

### LINKED ISSUE #2108: Python interface
Build a python interface, using e.g. [hyphen](https://github.com/tbarnetlamb/hyphen). Start with the core #2107 API #1790. The python interface could be used to interface Hets with https://github.com/gavel-tool and to build new interfaces, like a new GUI.

b-gehrke:
Basic interface for the python API. Contains
- haskell modules for wrapping existential data types and other non expressible haskell features into function callable from python 
- haskell modules grouping required functions and datatypes in few modules to allow performant loading
- python type annotations for haskell functions
- python classes wrapping around haskell functions allowing a python-like access

Closes #2108


tillmo (changes requested):
Generally, all Hets modules (both Haskell and Python) should start with a
header that explains the purpose of the module (see other Hets modules).

I do not understand why the duplication in ProofOptions and PyProofOptions
is needed. For example, G_prover is an existential datatype. But in the
datatype ProofOptions, G_prover is wrapped with a Maybe. Wouldn't this
suffice in order to hide the existential type?

python/hets/DevGraphEdge.py
```diff
@@ -0,0 +1,26 @@
+from typing import Tuple, Optional
```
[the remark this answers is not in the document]
b-gehrke: Done, added as `morphism`

HetsAPI/DataTypes.hs
```diff
@@ -0,0 +1,30 @@
...
+import Data.Graph.Inductive.Graph (LNode)
+
+-- | In the API if it is not possible to export the generic type, they are converted to JSON.
+type GenericTransportType = ByteString
+
+type Sentence = GenericTransportType
+type SentenceByName = OMap.OMap String Sentence
+
+type SignatureJSON = GenericTransportType
+type SymbolJSON = GenericTransportType
```
tillmo, line 28: This seems to be problematic, because sentences can contain UTF-8 characters. Why not Data.Text?

python/hets/json_conversion.py
```diff
@@ -0,0 +1,23 @@
...
+    Data.Aeson.
+    See https://hackage.haskell.org/package/aeson/docs/Data-Aeson.html
+    for further details.
+    """
+
+    # Get json.
+    # Return type is ByteString which is not converted to either pythons
+    # `bytes` or `str` type. Hence, the conversion via `show` and
+    # `encode.decode`. [1:-1] to exclude the quotation marks added by `show`
+    json_str = show(hs_byte_string).encode("utf-8").decode("unicode_escape")[1:-1]
```
tillmo, line 20: why `unicode_escape`?

HetsAPI/Python.hs
```diff
@@ -86,49 +223,148 @@ instance Show PyProver where
...
+comorphismOfGMorphism :: PyGMorphism -> PyComorphism
+comorphismOfGMorphism (PyGMorphism (GMorphism {gMorphismComor = cid})) = PyComorphism (Comorphism cid)
+
+signatureOfGMorphism :: PyGMorphism -> ExtSign HDT.SignatureJSON HDT.SymbolJSON
+signatureOfGMorphism (PyGMorphism (GMorphism {gMorphismSign = sig})) = ExtSign {
+        plainSign = encode (plainSign sig),
+        nonImportedSymbols = Set.map encode $ nonImportedSymbols sig
+    }
+
+logicNameOfGMorphism :: PyGMorphism -> String
```
tillmo, line 349: this is not the logic name, but the comorphism name

tillmo (changes requested):

python/hets/Sentence.py
```diff
@@ -0,0 +1,19 @@
+from typing import Tuple, Callable
```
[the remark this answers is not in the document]
b-gehrke: Function for json conversion is extracted to json_conversion.py and documented there.

HetsAPI/DataTypes.hs
b-gehrke: Why is it problematic? There is no problem storing UTF8 characters in a ByteString. ByteString is used because Aeson serializes to a ByteString.

python/hets/json_conversion.py
b-gehrke: In the future, it might be a good idea to extend hyphen to convert haskell `Data.ByteString(.Lazy).ByteString` to pythons `bytes`.

HetsAPI/Python.hs
b-gehrke: Done, changed to `comorphismNameOfGMorphism` and `comorphismDescriptionOfGMorphism`.

python/hets/json_conversion.py
b-gehrke: Changed to decode `ByteString` directly.

tillmo (approved):

PR #2169, opened 2026-01-05
Add a copy of `LICENSE.txt` to `atermlib`

neilmayhew:
This is necessary to allow `atermlib` to be built standalone (eg from Hackage) because `atermlib.cabal` refers to it.


neilmayhew:
A new Hackage release would be appreciated. The last one was in 2020, and the package is currently broken.

